PR #11, opened 2022-08-23
feat: support ipwidgets v8

maartenbreddels:
Although in the frontend we still use ipywidgets v7, most thing we do
do not care about this.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -19,6 +19,7 @@ jobs:
         os: [ubuntu, windows]
         # just 1 version, it's heavy
         python-version: [3.8]
+        ipywidgets: ["7.7", "8.0"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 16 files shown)


maartenbreddels:
I just noticed this today, because i was fixing ipyvolume to work in ipywidgets. Let me see if I can rebase this PR, and see if I can find the issue.

maartenbreddels:
Note, this also requires:
```
$ (cd packages/solara-widget-manager; npm instal; npm run build)
$ (cd packages/solara-vuetify-app; npm instal; npm run build; npm run devlink)
```

maartenbreddels:
Make sure to clear your browser cache, because if we don't change the version number during dev, the browser will pick up the cached version (Dev console -> Network -> Disable cache in chrome)

mangecoeur:
Thanks, I give this a go on Friday but I didn't quite get it working even following your instructions, but that might have been more end-of-the week brain fade problems ;)

PR #16, opened 2022-10-21
refactor: use the comm package instead of ipykernel

maartenbreddels:
This will avoid using a HasTraits object (faster)
and removed more ipykernel dependency.

cc @martinRenou


maartenbreddels:
superseded by https://github.com/widgetti/solara/pull/22

PR #29, opened 2023-01-27
Add docs for reading data into pandas df

Ben-Epstein:


solara/components/file_drop.py
```diff
@@ -58,6 +58,24 @@ class FileInfo(typing.TypedDict):
...
+    import pandas as pd
+    import solara
+
+    @solara.component
+    def Page():
+        def load_file_df(file):
+            data = io.BytesIO(file["data"])
+            display(pd.read_csv(data))
+        
+        solara.FileDrop(label="Drop file to see dataframe!",lazy=False, on_file=load_file_df)
```
maartenbreddels, line 76:
```suggestion
    @solara.component
    def Page():
        def load_file_df(file):
            display(pd.read_csv(file["file_obj"]))
        
        solara.FileDrop(label="Drop file to see dataframe!",lazy=True, on_file=load_file_df)
```
I think this will work, right @mariobuikhuizen ?
No need to download it first.

maartenbreddels: display(..) only works in Jupyter btw. We do plan to support it, but an example with set_df(...) would be more appropriate.

maartenbreddels (commented):
Awesome to see your first PR here Ben!

solara/components/file_drop.py
Ben-Epstein: I will test if that works, I thought it didn't in my application

Ben-Epstein: @maartenbreddels that does not work at the moment (which is why I ended up on the path to wanting to document it :) )

I get this error

```
    new_df = pd.read_csv(file["file_obj"])
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/util/_decorators.py", line 211, in wrapper
    return func(*args, **kwargs)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/util/_decorators.py", line 331, in wrapper
    return func(*args, **kwargs)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 950, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 605, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1442, in __init__
    self._engine = self._make_engine(f, self.engine)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/io/parsers/readers.py", line 1753, in _make_engine
    return mapping[engine](f, **self.options)
  File "bulk-labeling-solara/.venv/lib/python3.9/site-packages/pandas/io/parsers/c_parser_wrapper.py", line 79, in __init__
    self._reader = parsers.TextReader(src, **kwds)
  File "pandas/_libs/parsers.pyx", line 554, in pandas._libs.parsers.TextReader.__cinit__
pandas.errors.EmptyDataError: No columns to parse from file

```

maartenbreddels:
I pushed what works here locally, can you try this out?

solara/components/file_drop.py
maartenbreddels: FYI, the next version of Reacton (which Solara will use) will support display, although it wouldn't work in the callback.

maartenbreddels (approved):

iisakkirotko:
Thanks for the contribution @Ben-Epstein, we finally got it in! :)

PR #30, opened 2023-02-06
feat: Download component to allow users to download data to a file

maartenbreddels:
cc @Ben-Epstein


Ben-Epstein (approved):

maartenbreddels:
Renamed to FileDownload, and we should rename/change FileDrop to FileUpload for consistent naming and easy to find components.

PR #31, opened 2023-02-16
feat: add option for a custom widget in column menu

mariobuikhuizen:


solara/components/datatable.py
```diff
@@ -19,7 +19,17 @@ def format_default(df, column, row_index, value):
...
+def DataTable(
+    df,
+    page=0,
+    items_per_page=20,
+    format=None,
+    column_actions: List[ColumnAction] = [],
+    cell_actions: List[CellAction] = [],
+    scrollable=False,
+    on_column_menu_open: Optional[Callable[[Optional[str]], None]] = None,
+    column_widget=None,
```
maartenbreddels, line 31:
```suggestion
    column_menu_element : Optional[reacton.core.Element] = None,
```
If used from a component, it should be a element, in the widget layer it will indeed be translated to a widget. And for symmetry with `on_column_menu_open`, we can call this the column_menu_element I think.

solara/components/datatable.py
```diff
@@ -69,6 +79,8 @@ def DataTable(df, page=0, items_per_page=20, format=None, column_actions: List[C
         cell_actions=cell_actions,
         _column_actions_callbacks=column_actions_callbacks,
         _cell_actions_callbacks=cell_actions_callbacks,
+        on_column_menu_open=on_column_menu_open,
+        column_widget=column_widget,
```
maartenbreddels, line 83:
```suggestion
        column_menu_widget=column_menu_element,
```

maartenbreddels (commented):
Nice. But after reviewing.. I think `column_header_widget`  / `column_header_open` is a better name. What do you think?

PR #34, opened 2023-03-08
feat: query parameter hook

maartenbreddels:
Allows to use query parameters to manage state.

### DIFF
--- a/solara/__init__.py
+++ b/solara/__init__.py
@@ -37,6 +37,7 @@ from reacton import (
 )  # noqa: F403, F401
 from reacton.core import Element  # noqa: F403, F401
 from . import util
+from reacton.utils import equals
 
 # flake8: noqa: F402
 from .datatypes import *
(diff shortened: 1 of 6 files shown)


maartenbreddels:
@koenvo this is what I planned for the query parameter API.

But with solara 1.7.1 you can use the more crude API:
```
    ....
    router = use_router()
    router.push("/some/path?count=10")
```

koenvo:
> @koenvo this is what I planned for the query parameter API.
> 
> But with solara 1.7.1 you can use the more crude API:
> 
> ```
>     ....
>     router = use_router()
>     router.push("/some/path?count=10")
> ```

Great! For me it’s most important to be able to access the query parameters instead of pushing new ones. Should that already work using use_router().search?

maartenbreddels:
> Great! For me it’s most important to be able to access the query parameters instead of pushing new ones. Should that already work using use_router().search?

Indeed, that should be enough for your use case, so you don't need to wait til this branch is merged and released.

Issue #36, opened 2023-03-09
Quickstart example does not render correctly with solara run


Jhsmit:
I have a fresh conda environment with python 3.9 on windows. I've used poetry to install solara.

I have a `myapp.py` file with the following code:
```python
import solara

@solara.component
def Page():
    clicks, set_clicks = solara.use_state(0)
    def increase_clicks():
        set_clicks(clicks+1)
    solara.Button(label=f"Clicked {clicks} times", on_click=increase_clicks)
```

And then run with:
```
solara run `myapp.py`
```

The webpage looks like this:
![image](https://user-images.githubusercontent.com/7881506/223989473-bfa52927-e7b5-4c11-ac6d-6df7b3fad347.png)

These are some of the errors in the console:
```
The stylesheet http://localhost:8765/_solara/cdn/font-awesome@4.5.0/css/font-awesome.min.css was not loaded because its MIME type, “text/plain”, is not “text/css”.
The script from “http://localhost:8765/_solara/cdn/mermaid@9.1.7/dist/mermaid.min.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type
The script from “http://localhost:8765/_solara/cdn/requirejs@2.3.6/require.js” was loaded even though its MIME type (“text/plain”) is not a valid JavaScript MIME type.
The stylesheet http://localhost:8765/_solara/cdn/@widgetti/solara-vuetify-app@3.0.1/dist/main.css was not loaded because its MIME type, “text/plain”, is not “text/css”.
```

maartenbreddels:
Which browser are you using?

maartenbreddels:
And could you check in the browser dev console, what the response is to `http://localhost:8765/_solara/cdn/font-awesome@4.5.0/css/font-awesome.min.css`?

Of if you have curl, what is the output of
```
$ curl -i http://localhost:8765/_solara/cdn/font-awesome@4.5.0/css/font-awesome.min.css
```

Jhsmit:
I'm using firefox but I have the same issue with chrome. 

I've now tried a clean conda env with `pip install solara` which gives the same result. 
the curl output (I run this from another cmd.exe console): 

```
HTTP/1.1 200 OK
date: Thu, 09 Mar 2023 10:51:25 GMT
server: uvicorn
content-type: text/plain; charset=utf-8
content-length: 27466
last-modified: Thu, 09 Mar 2023 10:48:34 GMT
[...]

PR #37, opened 2023-03-09
fix: get correct mime types on windows for cdn proxy.

maartenbreddels:
Because on windows we cached to a file ending in /path.css/__main The mime type was not guessed correctly.

Fixes https://github.com/widgetti/solara/issues/36

### DIFF
--- a/solara/server/cdn_helper.py
+++ b/solara/server/cdn_helper.py
@@ -52,9 +52,9 @@ def get_data(base_cache_dir: pathlib.Path, path):
         raise Exception(f"Could not load URL: {url}")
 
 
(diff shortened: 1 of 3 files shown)


maartenbreddels:
It should fix it indeed. Let me check on windows manually again.

maartenbreddels:
Ok, mystery solved. You have to nuke your cdn cache directory, e.g. sth like this:
```
$ rm -rf /c/ProgramData/Miniconda3/envs/solara39/share/solara/cdn/
```

Jhsmit:
image

🎉 🥳

PR #40, opened 2023-03-10
feat: use_reactive as alternative for use_state

maartenbreddels:
Many people have trouble digesting `use_state`. This could be an alternative way of using state in solara. Needs user testing.


maartenbreddels:
superseded by https://github.com/widgetti/solara/pull/72

PR #42, opened 2023-03-15
feat: Reactive variables will automatically re-execute a component

maartenbreddels:
when accessed. With this, we don't need to call .use() anymore.

Requires https://github.com/widgetti/reacton/commit/5fe6085fcc156cb76fcb1c709d0d1540b76f1ee9 to be released.


maartenbreddels:
@Ben-Epstein
You might like this:
image

PR #43, opened 2023-03-16
fix: docstring typo

Jhsmit:


maartenbreddels:
So, are you saying solara is not important? :)

PR #46, opened 2023-03-23
feat: Add FileInput component

Jhsmit:


solara/components/file_input.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # workaround for @observe being cleared
+        real.version += 1
+        real.reset_stats()
+
+        set_wired_files(cast(List[FileInfo], real.get_files()))
+
+    solara.use_effect(wire_files, [file_info])
+
+    def handle_file(cancel: threading.Event) -> None:
```
maartenbreddels, line 78:
```suggestion
    def handle_file() -> None:
```

solara/components/file_input.py
```diff
@@ -0,0 +1,96 @@
+"""
+FileInput component.
+"""
+
+
+import threading
```
maartenbreddels, line 6:
```suggestion
```

solara/components/file_input.py
```diff
@@ -0,0 +1,96 @@
...
+import solara
+import solara.hooks as hooks
+from solara.components.file_drop import FileInfo
+
+
+@solara.component
+def FileInput(
+    label: str = "",
+    on_total_progress: Optional[Callable[[float], None]] = None,
+    on_file: Optional[Callable[[List[FileInfo]], None]] = None,
```
maartenbreddels, line 20:
```suggestion
    on_file: Union[None, Callable[[FileInfo], None], Callable[[List[FileInfo]], None]] = None,
```
lets make the case for multiple=False simpler

solara/components/file_input.py
```diff
@@ -0,0 +1,96 @@
...
+    accept: str = "",
+    multiple: bool = True,
```
maartenbreddels, line 22:
```suggestion
    multiple: bool = False,
```
I think this is a better default, will make the default behaviour similar to FileDrop

solara/components/file_input.py
maartenbreddels: this is not needed anymore

solara/components/file_input.py
```diff
@@ -0,0 +1,96 @@
...
+        if not wired_files:
+            on_file([])
+        if lazy:
+            for f_info in wired_files:
+                f_info["data"] = None
+        else:
+            for f_info in wired_files:
+                f_info["data"] = f_info["file_obj"].read()
+
+        on_file(wired_files)
```
maartenbreddels, line 90:
```suggestion
        if multiple:
                    on_file(wired_files)
        else:
                    on_file(wired_files[0])
  
```

maartenbreddels (changes requested):
Nice work, many thanks!

Looks good. Some minor comments/changes

cc @mariobuikhuizen @Ben-Epstein

I do wonder if we need to have some integration between FileInput and FileDrop. Maybe FileInput should allow also a drop action?
@mariobuikhuizen do you know if this is technically possible?

solara/components/file_input.py
```diff
@@ -0,0 +1,103 @@
...
+
+from ipyvuetify.extra import FileInput as ExtraFileInput
+
+import solara
+import solara.hooks as hooks
+from solara.components.file_drop import FileInfo
+
+
+@solara.component
+def FileInput(
```
maartenbreddels, line 16:
```suggestion
def InputFile(
```
We have postfixed all input components with Input, such as InputText/Int/Float, so I think it makes sense to rename it like this.

Jhsmit:
This is a nice one, a drop zone which you can also click to 

image
(from chatpdf.com)

Thats looks like to more or less already exist / be easy to do
https://codepen.io/hkchen/pen/EraZXm

lp9052:
Any updates about this PR? This one provides a more comprehensive file input option than Filedrop, especially the accept and disable keyword. I see you did not explicitly add disable keyword in the param list. Do you want to add it? I do believe it will supported by kwarg, since disable is a common and useful keyword, maybe you want to directly add to the signature?

maartenbreddels:
I've taken a quick look/pass on this PR, and there are some issues
 * ipyvuetify does not pass class_, style_ and label to the vuetify component in the template: https://github.com/widgetti/ipyvuetify/blob/master/ipyvuetify/extra/file_input.vue
 * on_file triggers directly with None
 * better typing for multiple=True/False
 * integration test needed.

Hope to do another pass soon, but if someone want to pick up some items, that would be great!

lp9052:
> I've taken a quick look/pass on this PR, and there are some issues
> 
> * ipyvuetify does not pass class_, style_ and label to the vuetify component in the template: https://github.com/widgetti/ipyvuetify/blob/master/ipyvuetify/extra/file_input.vue
> * on_file triggers directly with None
> * better typing for multiple=True/False
> * integration test needed.
> 
> Hope to do another pass soon, but if someone want to pick up some items, that would be great!

@maartenbreddels Do you have other tickets/feature you need help? If not, I can take a look at this.

maartenbreddels:
Those would be good to work on, otherwise see https://github.com/widgetti/solara/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22 
Would love to see more contributions from you 😍

PR #59, opened 2023-04-07
feat: Allow FileBrowser to show folders first

Jhsmit:
This PR adds a boolean `folders_first` to the `FileBrowser`, which makes folders appear first in the sorting.

Also fixes a bug where the default filter for `list_dir` was the default built-in `filter`

### DIFF
--- a/solara/components/file_browser.py
+++ b/solara/components/file_browser.py
@@ -18,7 +18,7 @@ def list_dir(path, filter: Callable[[Path], bool] = lambda x: True, folders_firs
         return {"name": n, "is_file": is_file, "size": humanize.naturalsize(os.stat(full_path).st_size) if is_file else None}
 
     def keyfunc(item):
-        if item['is_file']:
+        if item["is_file"]:
             return str(int(folders_first)) + item["name"].lower()
         else:
             return str(int(not folders_first)) + item["name"].lower()
--- a/solara/components/file_browser.py
+++ b/solara/components/file_browser.py
@@ -11,14 +11,21 @@ import solara
 from solara.components import Div
 
 
-def list_dir(path, filter: Callable[[Path], bool] = lambda x: True):
+def list_dir(path, filter: Callable[[Path], bool] = lambda x: True, folders_first: bool = False) -> List[dict]:
     def mk_item(n):
         full_path = join(path, n)
         is_file = isfile(full_path)
         return {"name": n, "is_file": is_file, "size": humanize.naturalsize(os.stat(full_path).st_size) if is_file else None}
 
+    def keyfunc(item):
+        if item['is_file']:
+            return str(int(folders_first)) + item["name"].lower()
+        else:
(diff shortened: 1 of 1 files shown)


solara/components/file_browser.py
```diff
@@ -11,14 +11,21 @@
 from solara.components import Div
 
 
-def list_dir(path, filter=filter):
+def list_dir(path, filter: Callable[[Path], bool] = lambda x: True, folders_first: bool = False) -> List[dict]:
```
maartenbreddels, line 14:
```suggestion
def list_dir(path, filter: Callable[[Path], bool] = lambda x: True, directory_first: bool = False) -> List[dict]:
```

solara/components/file_browser.py
```diff
@@ -11,14 +11,21 @@
...
     def mk_item(n):
         full_path = join(path, n)
         is_file = isfile(full_path)
         return {"name": n, "is_file": is_file, "size": humanize.naturalsize(os.stat(full_path).st_size) if is_file else None}
 
+    def keyfunc(item):
+        if item["is_file"]:
+            return str(int(folders_first)) + item["name"].lower()
```
maartenbreddels, line 22:
```suggestion
            return (folders_first, item["name"].lower())
```

solara/components/file_browser.py
```diff
@@ -11,14 +11,21 @@
...
+        else:
+            return str(int(not folders_first)) + item["name"].lower()
```
maartenbreddels, line 24:
```suggestion
            return (not folders_first, item["name"].lower())
```

solara/components/file_browser.py
```diff
@@ -78,6 +86,7 @@ def FileBrowser(
      * `on_path_select`: Depends on mode, see above.
      * `on_file_open`: Depends on mode, see above.
      * `filter`: A function that takes a `Path` and returns `True` if the file/directory should be shown.
+     * `folders_first`: If `True` folders are shown before files. Default: `False`.
```
maartenbreddels, line 89:
```suggestion
     * `directory_first`: If `True` directories are shown before files. Default: `False`.
```

maartenbreddels (changes requested):
Nice, thank you!
I think it's better we sort tuples (not a related thing, but lets do it right directly).
Also, we choose to use 'directory' instead of 'folders'
Hope you agree

Jhsmit:
I agree, I tried tuples for sorting but it didnt works as I expected. 
But now it seems to work and it makes it much cleaner. 

Question: should the default value for `directory_first` be `True`, since that is the default on windows?
Also it is my preferred behavior, but if we change it its a 'breaking change'

maartenbreddels:
Yeah, it would be breaking behaviour. Lets keep it as it is.

maartenbreddels (approved):

maartenbreddels:
Thank you Jochem!

PR #71, opened 2023-04-18
fix proplot import

Jhsmit:
`proplot` configures in inline plotting backend on import with ipython magic.
https://github.com/proplot-dev/proplot/blob/d44b25b39f3065d9f603d13416fa5b6c22feb1dd/proplot/config.py#L361

AFAIK the configuration settings does not affect anything when using proplot with solara, but solara's  `FakeIPython` needs to have a fake `magic` method for proplot to be imported. 

Meanwhile, users can monkey patch `FakeIPython`

```python
from solara.server.patch import FakeIPython
FakeIPython.magic = lambda *args: None
```

### DIFF
--- a/solara/server/patch.py
+++ b/solara/server/patch.py
@@ -64,6 +64,9 @@ class FakeIPython:
                 # TODO: should we remove it from the list?
                 pass
 
+    def magic(self, *args):
+        # proplot requires this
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thank you Jochem!

PR #73, opened 2023-04-21
chore: better ci, make workflows reusable

maartenbreddels:


maartenbreddels:
Inspiration from @12rambau https://github.com/widgetti/ipyvuetify/issues/257#issuecomment-1517659888

maartenbreddels:
Not ideal, since it will build multiple times, at least the code/yaml is not duplicated

PR #74, opened 2023-04-26
docs: remove `cancelled` arg from work function

Jhsmit:
Assuming this is no longer needed

### DIFF
--- a/solara/website/pages/api/use_thread.py
+++ b/solara/website/pages/api/use_thread.py
@@ -1,4 +1,3 @@
-import threading
 import time
 from pathlib import Path
 from typing import Optional, cast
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thanks, always embarrassing when CI fails on an unrelated issue :)

Jhsmit:
hmm isnt this setting supposed to fix that?
Otherwise all external PRs would always fail
image

maartenbreddels:
Where do you see that setting?

Jhsmit:
I think only I can see it as contributer of a PR
But it looks like it reveals my secret rather than yours
Still, wouldnt it make sense that since we are on widgetti/solara that the PR has access to GitHub secrets?

maartenbreddels:
No, because that would allow anyone to add `echo $SOME_SECRET` to our CI, and steal our secrets :)

maartenbreddels:
Sorry for all the noise, and thanks you for testing our CI pipeline for PR's :)

PR #76, opened 2023-05-03
fix: do not call use_reactive on_value when value is updated externally

maartenbreddels:
It does not make sense to trigger on_value when the value passed to use_reactive is changed. This has the side effect of the checkbox' on_value being triggered when the checkbox' value is changed by the code (it should only be triggered when the user clicks the checkbox, or the reactive value is changed by some other part of the code).

@Jhsmit could you test and review this?

### DIFF
--- a/tests/unit/checkbox_test.py
+++ b/tests/unit/checkbox_test.py
@@ -1,0 +1,29 @@
+from unittest.mock import MagicMock
+
+import ipyvuetify as vw
+
+import solara
+
+
+def test_checkbox():
+    on_value = MagicMock()
+    el = solara.Checkbox(label="label", value=True, on_value=on_value)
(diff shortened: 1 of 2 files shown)


tests/unit/checkbox_test.py
```diff
@@ -0,0 +1,29 @@
...
+    rc.close()
+
+
+def test_checkbox_no_callback_on_managed():
+    on_value = MagicMock()
+    el = solara.Checkbox(label="label", value=True, on_value=on_value)
+    box, rc = solara.render(el, handle_error=False)
+    assert on_value.call_count == 0
+
+    # chaning the value should call the callback
```
Jhsmit, line 26: Should _not_ call the callback?

Jhsmit:
This now fixes my issue and passes my tests

PR #85, opened 2023-05-12
docs: fix modify script code suggestion

salfaris:
Currently, the code suggestion for the modify script is incorrect. The initial `sol.py` does not use a return a statement. It also use `increment` instead of `on_click` as the event handler. Also, to check for `clicks == 0`, we need to use the state value `clicks.value == 0` instead. Otherwise, nothing happens. I assume this is just a mismatch due to a recent API change.


maartenbreddels:
Thanks a lot. Sorry we missed this during the refactor! But happy to find out you were able to fix it!

maartenbreddels:
Thank you for your contribution!

PR #102, opened 2023-05-19
Make sure to set the correct cookie attributes when serving from loca…

koenvo:
…lhost

### DIFF
--- a/solara/server/flask.py
+++ b/solara/server/flask.py
@@ -5,6 +5,7 @@ import os
 from http.server import HTTPServer
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlparse
 from uuid import uuid4
 
 import flask
@@ -220,9 +221,20 @@ def read_root(path):
     if not allowed():
         return login()
 
+    samesite = "lax"
+    secure = False
+    # we want samesite, so we can set a cookie when embedded in an iframe, such as on huggingface
+    # however, samesite=none requires Secure https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+    # when hosted on the localhost domain we can always set the Secure flag
+    # to allow samesite https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies
+    o = urlparse(request.base_url)
+    if request.headers.get("x-forwarded-proto", "http") == "https" or o.hostname == "localhost":
+        samesite = "none"
+        secure = True
+
     assert session_id is not None
     response = flask.Response(content, mimetype="text/html")
-    response.set_cookie(server.COOKIE_KEY_SESSION_ID, value=session_id)
+    response.set_cookie(server.COOKIE_KEY_SESSION_ID, value=session_id, secure=secure, samesite=samesite)
     return response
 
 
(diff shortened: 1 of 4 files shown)


solara/server/flask.py
```diff
@@ -220,9 +221,20 @@ def read_root(path):
     if not allowed():
         return login()
 
+    samesite = "lax"
```
koenvo, line 224: I added some duplicate code here. What would be a good place to store this, if you like it to be refactored. Other option might be to test only the `starlette` server and don't touch the `flask` part.

maartenbreddels: It’s ok, since we test it, unless you see an elegant way

maartenbreddels:
Very cool, and high quality pr 😍
Do the test really test it since it is the same server and port? If you revert the changes, does the test fail?

koenvo:
> Very cool, and high quality pr 😍 Do the test really test it since it is the same server and port? If you revert the changes, does the test fail?

You're welcome. Happy to contribute!


Good call here. Normally I would write the failing test first 🙃  

Output with the changes reverted:
```bash
FAILED tests/integration/server_test.py::test_run_in_iframe[flask-chromium] - playwright._impl._api_types.TimeoutError: Timeout 18000ms exceeded.
FAILED tests/integration/server_test.py::test_run_in_iframe[starlette-chromium] - playwright._impl._api_types.TimeoutError: Timeout 18000ms exceeded.
```

maartenbreddels:
I did some local testing, and managed to pass the test after disabling your feature. I think it is because it is running on the same server.

Then I made it run on a different port, and it still passes the test (I can see with --headed and PWDEBUG=1 that the page really loads in the iframe). If I test this with `python http.server` like in the docs, it *does* fail.

I think there is still something happening I don't understand.

maartenbreddels:
e.g. try changing localhost to xlocalhost in the check of starlette and flask and the test pass 🤷

maartenbreddels:
Ah, we can use `    page_session.context.clear_cookies()` for this.



> 2\. `http://localhost` are considered cross-origins which is good for the test.

I didn't know! Where did you find this?


Ok, I can confirm that now with clearing the cookies, it fails without your changes.

koenvo:
> Ah, we can use ` page_session.context.clear_cookies()` for this.
Right! I was thinking about restarting the entire webserver but this is what I was actually looking for.

> 
> > 2. `http://localhost` are considered cross-origins which is good for the test.
> 
> I didn't know! Where did you find this?
> 
Test it and use `PWDEBUG` to see if the cookies get accepted. By default the `page_session.url` is `about:blank`

tests/integration/server_test.py
```diff
@@ -124,3 +124,34 @@ def test_from_thread_two_users(browser: playwright.sync_api.Browser, solara_serv
...
         page2.close()
+
+
+@solara.component
+def StaticPage():
+    return w.Label(value="Hello world")
+
+
+def test_run_in_iframe(page_session: playwright.sync_api.Page, solara_server, solara_app, extra_include_path, tmpdir):
+    tmpdir.join("index.html").write(
```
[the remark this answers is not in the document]
maartenbreddels: thanks!

maartenbreddels:
Great work, thanks Koen!

PR #111, opened 2023-05-23
correct call to run portal

Dorianux:
updated call to run created portal 
(was mentioned correctly above)


maartenbreddels:
Thank you! I thought I already merged this, sorry for taking to long.

PR #137, opened 2023-06-02
Fix typos

giswqs:
This PR fixes many typos detected by [codespell](https://github.com/codespell-project/codespell). The changes only affect plain text. No Python source code has been changed.

### DIFF
--- a/.github/workflows/codequality.yaml
+++ b/.github/workflows/codequality.yaml
@@ -23,8 +23,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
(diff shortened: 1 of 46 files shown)


maartenbreddels:
Thank you! I felt like i should have done this a long time ago. I've now added this to GHA

giswqs:
There are a couple more errors in some files with non-English text in it. I did not correct them. You might need to exclude those words in codespell. 

image

giswqs:
Something like this if those words are not errors:

```bash
codespell --ignore-words-list="convertor,prolem,ore"
```

maartenbreddels:
Thanks for starting this!

PR #139, opened 2023-06-03
Minor change in wording also fixing a typo

deeplook:


maartenbreddels:
Thank you. I don't think we really use this file anymore. I think we should delete it. Do you want to do that?

deeplook:
It's used here here in the Page class. Should that vanish, too? https://github.com/widgetti/solara/blob/master/solara/website/pages/__init__.py#L13

maartenbreddels:
Best is to replace it by this for the moment. 
```python
@solara.component
def Page():
    solara.Markdown("should not see me")
```

deeplook:
Maybe that's not enough with an integration test giving this: `waiting for locator("text=Test in solara") to be visible`.

maartenbreddels:
Thank you!

Issue #78, opened 2023-05-04
Support latex


maartenbreddels:
We currently don't support latex. Although we can support it easily on initial page load, we should also make sure they get updated when the widget value updates.
See https://github.com/voila-dashboards/voila/pull/531 for inspiration

PR #155, opened 2023-06-15
Fix: latex support

maartenbreddels:
We now re-render latex as I proposed in https://github.com/voila-dashboards/voila/pull/531 my mocking the v2 API of MathJax.


maartenbreddels:
Closes https://github.com/widgetti/solara/issues/78

Issue #112, opened 2023-05-24
Running solara app in a docker container


legout:
Hi guys,

I am not able to run my solara app in a docker container. Running the docker container results in the following error:

`ERROR:    [Errno 99] error while attempting to bind on address ('::1', 8765, 0, 0): cannot assign requested address`

Here is what my Dockerfile looks like
```Dockerfile
#
# Build image
#

FROM python:3.11-slim as builder

ENV PYTHONFAULTHANDLER=1 \
    PYTHONHASHSEED=random \
    PYTHONUNBUFFERED=1 \
    POETRY_HOME=/opt/poetry

WORKDIR /app
COPY . .

RUN apt update -y && apt upgrade -y && apt install curl -y
RUN curl -sSL https://install.python-poetry.org | python3 -
RUN ${POETRY_HOME}/bin/poetry config virtualenvs.create false 
RUN ${POETRY_HOME}/bin/poetry install --no-dev 
RUN ${POETRY_HOME}/bin/poetry export -f requirements.txt >> requirements.txt


#
# Prod image
#

FROM python:3.11-slim AS runtime


WORKDIR /app
COPY . .
COPY --from=builder /app/requirements.txt .
RUN pip install --no-cache-dir -r requirements.txt


EXPOSE 8765
CMD ["solara", "run", "src/app.py"]

```

legout:
Solved the problem by adding `--host=0.0.0.0` to `solara run` in the Dockerfile.

```Dockerfile

...

CMD ["solara", "run", "src/app.py", "--host=0.0.0.0"]
```

Marsh-sudo:
Would like to try solving this if possible

maartenbreddels:
Great, looking forward to hearing about it. Even if you don’t solve it right away, please  report what you found!

rmorshea:
I'm pretty sure [this is expected](https://stackoverflow.com/a/59182290) and not a problem with Solara.

maartenbreddels:
Hi Ryan, thanks for chiming in.

I don't think the issue in that link is what is happening here (although very important).
I think here the `::1` that we see in the error message comes from IPV6, which is probably used in combination with https://docs.python.org/3/library/socket.html#socket.AF_INET (and not AF_INET6)

Marsh-sudo:
> Thanks for reporting this and giving a solution. It would be nice if solara would work without this, so I'll keep it open for now.

So would you like the app in the container to run without ipv6 or with it so i can give a straight forward answer kindly

[...]

PR #167, opened 2023-06-22
Allow test host to be customized

astrofrog:
I ran into the same issue as https://github.com/widgetti/solara/issues/112 but when running solara though the pytest framework which means I wasn't able to set the host to 0.0.0.0. This PR adds a way to customize the host with a new environment variable. I thought HOST would be too generic so made it SOLARA_TEST_HOST. I wonder if PORT should be SOLARA_TEST_PORT too?

### DIFF
--- a/solara/website/pages/docs/content/10-howto/50-testing.md
+++ b/solara/website/pages/docs/content/10-howto/50-testing.md
@@ -131,3 +131,13 @@ After inspecting and approving the screenshots, you can copy them to the `solara
 Visual testing with solara is based on [Playwright for Python](https://playwright.dev/python/), which provides a `page` fixture. However, this fixture will make a new page for each test, which is not what we want. Therefore, we provide a `page_session` fixture that will reuse the same page for all tests. This is important because it will make the tests faster.
 
 By following these recommendations and guidelines, you can efficiently test your Solara applications and ensure a smooth developer experience.
+
+# Configuration
+
+## Changing the Hostname
+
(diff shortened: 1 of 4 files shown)


solara/website/pages/docs/content/10-howto/50-testing.md
```diff
@@ -131,3 +131,13 @@ After inspecting and approving the screenshots, you can copy them to the `solara
 Visual testing with solara is based on [Playwright for Python](https://playwright.dev/python/), which provides a `page` fixture. However, this fixture will make a new page for each test, which is not what we want. Therefore, we provide a `page_session` fixture that will reuse the same page for all tests. This is important because it will make the tests faster.
 
 By following these recommendations and guidelines, you can efficiently test your Solara applications and ensure a smooth developer experience.
+
+# Configuration
+
+## Changing the Hostname
+
+To configure the hostname the socket is bound to when starting the test server, use the `HOST` or `SOLARA_HOST` environment variable (e.g. `SOLARA_HOST=0.0.0.0`). This hostname is also used for the jupyter server and voila. Alternatively the `--solara-host` argument can be passed on the command line for pytest.
```
astrofrog, line 139: Is SOLARA_HOST actually recognized currently?

maartenbreddels: No, I've added that in this PR

maartenbreddels: I think that should be good enough for when HOST is not specific enough, and collides with other usage. And in the case where even that is not enough, `--solara-host` can be used for the testing host. I think that provides enough configuration/escape hatches.

Issue #144, opened 2023-06-05
[Enhancement] Select widget missing keyword disabled


lp9052:
Select and SelectMultiple widgets seem missing keyword disabled.

Is it by design or still under implementation?

Is there any work around instead of using ipyvuetify directly.

thanks!

maartenbreddels:
Hi,

we still need to add it. Should be easy to add in https://github.com/widgetti/solara/blob/7ec459c4d9f8e65ede0f24301da695cb7bcff8d1/solara/components/select.py#L14

Or otherwise use that code as an example (ignore all the type stuff).

Hope that helps.

Regards,

Maarten

PR #173, opened 2023-06-23
feat: add disable keyword to select and select multiple widgets

lp9052:
add disabled keyword to select widgets
see #144

### DIFF
--- a/solara/components/select.py
+++ b/solara/components/select.py
@@ -101,7 +101,7 @@ def Select(
      * `values`: List of values to select from.
      * `on_value`: Callback to call when the value changes.
      * `dense`: Whether to use a denser style.
-     * `disabled`: Whether widget should allow user interact
+     * `disabled`: Whether the select widget allow user interaction
      * `classes`: List of CSS classes to apply to the select.
      * `style`: CSS style to apply to the select.
 
@@ -162,7 +162,7 @@ def SelectMultiple(
      * `all_values`: List of all values to select from.
      * `on_value`: Callback to call when the value changes.
      * `dense`: Whether to use a denser style.
-     * `disabled`: Whether widget should allow user interact
+     * `disabled`: Whether the select widget allow user interaction
      * `classes`: List of CSS classes to apply to the select.
      * `style`: CSS style to apply to the select.
     """
--- a/solara/components/select.py
+++ b/solara/components/select.py
@@ -17,6 +17,7 @@ def Select(
     value: None = ...,
     on_value: Optional[Callable[[Optional[T]], None]] = ...,
     dense: bool = ...,
+    disabled: bool = ...,
     classes: List[str] = [],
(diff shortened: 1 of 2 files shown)


lp9052:
first PR. Want to use this to to familiarize with the process. 

Let me know if there are specific rules for this project. 

I followed instruction from development page.

Once this merged, I'll be able to contribute more.

Happy Friday.

solara/components/select.py
```diff
@@ -96,6 +101,7 @@ def Page():
      * `values`: List of values to select from.
      * `on_value`: Callback to call when the value changes.
      * `dense`: Whether to use a denser style.
+     * `disabled`: Whether widget should allow user interact
```
maartenbreddels, line 104:
```suggestion
     * `disabled`: Whether the select allow user interaction
```

tests/unit/select_test.py
```diff
@@ -0,0 +1,47 @@
...
+def test_select():
+    """
+    test select widget
+    """
+    disabled = None
+    on_value = MagicMock()
+    on_value_multiple = MagicMock()
+
+    @solara.component
+    def under_test():
```
maartenbreddels, line 17:
```suggestion
    def Test():
```

tests/unit/select_test.py
```diff
@@ -0,0 +1,47 @@
...
+        nonlocal disabled
+        disabled = solara.use_reactive(False)
+        solara.Select(label="single", values=["test0", "test1", "test2"], on_value=on_value, disabled=disabled.value)
+        solara.SelectMultiple(label="multiple", values=[], all_values=["test0", "test1", "test2"], on_value=on_value_multiple, disabled=disabled.value)
+
+    _, rc = solara.render(under_test(), handle_error=False)
```
maartenbreddels, line 23:
```suggestion
    _, rc = solara.render(Test(), handle_error=False)
```

maartenbreddels (changes requested):
Nice, thank you!

Just a few suggestions. Also in components_test.py we also have a test_select. I think we can remove that one, this one is better.

lp9052:
Thanks Maarten!

I'll remove the test in component test.

And is there any specific reason we prefer Test() over under_test() or test()? As snakecase is the preferred convention. Pep8 will report otherwise.

solara/components/select.py
lp9052: will change

lp9052:
urrr. I don't see components_test.py. Can you point out to me where is it?

maartenbreddels:
I see it is never added to git 🙃 
But we use the CamelCase because components are special functions, they are lazy, get re-executed, and contain state, which makes them behave more like classes.

lp9052:
just merged changes and please take a look

maartenbreddels:
Thank you!

Issue #177, opened 2023-06-26
Reload + Polars not working


Lundez:
Hi!

Polars (DataFrame) is built using Rust-bindings and this doesn't seem to work with Solara "Auto-Reload" function.

After first reload I get `NameError: name 'PyDataFrame' is not defined`, I believe because it's not reloading the imports correctly.
Because Polars uses rust it's using PyO3 to bind into Rust.

It's hard to debug as I don't understand Solara in-depth yet.
This works in streamlit.

How to reproduce:
```python
import solara
import polars as pl

@solara.component
def Page():
    file, set_file = solara.use_state(None)
    solara.FileDrop(on_file=set_file, lazy=False)
    solara.Text(f"Hello World!")
    Page2(file)

@solara.component
def Page2(file):

    if file is not None:
        solara.Text(f"File")
        df = pl.read_csv(file["data"])
        solara.DataFrame(df.to_pandas())
    else:
        solara.Text("No file, please upload one.")
```

Change a line and try to load a CSV again and it'll fail.. :/
Works on first try after starting server

Additionally I get a warning `Warning: polars binary missing!` which seems to suggest Rust bindings failing.

maartenbreddels:
Hi Lundez,

thanks for opening the issue. This is related to https://github.com/widgetti/solara/issues/148
I hope to take a look at this soon, or at least provide a workaround.

Thank you,

Maarten

Lundez:
@maartenbreddels if I add `polars` to the ignored reloaded imports (`reload.py`), e.g.
```python
reload_modules = {
            k for k in set(sys.modules) - set(self.ignore_modules) if not (k.startswith("solara.server") or k.startswith("anyio") or k.startswith("pandas") or k.startswith("polars"))
        }
```

it works!

Lundez:
Just to keep things both trackable (GitHub) and in more private discussions (Discord) I share our discussion here.

Lundez
> @MaartenBreddels i found that if we don't reload polars everything works as expected. I see pandas is excluded in the list (along with more libs), as such I did the same with polars. Posted this in issue.
>
> I went further and was thinking, what modules do we wish to reload? Perhaps we can only reload modules from our cwd (i.e. sys.path[0])? Why would we want to reload dependencies?

Marten
> @Londet i have that way of reloading now locally, because I also ran into issues. I think that should be the default for reloading from now on. Great we reach similar conclusions 🙂 I'd like to compare to streamlit as well btw, because looking at their code, I think they also reload everything.

Issue #148, opened 2023-06-09
Hot-reload error: PyO3 modules may only be initialized once per interpreter process


jonmmease:
For an app that uses the `VegaFusionWidget`, the first app render works well, but after editing the file the hot-reload step fails with this Python error:

```
No such comm: 25f3ea8b-a865-4a0c-b55e-8db9e4613d42
ERROR:    Uncaught exception: Traceback (most recent call last):
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 220, in m
    return(method(self, *args, **kwargs))
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 773, in _handle_msg
    self.set_state(state)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 650, in set_state
    with self._lock_property(**sync_data), self.hold_trait_notifications():
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1502, in hold_trait_notifications
    self.notify_change(change)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/ipywidgets/widgets/widget.py", line 701, in notify_change
    super().notify_change(change)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1517, in notify_change
    return self._notify_observers(change)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/traitlets/traitlets.py", line 1564, in _notify_observers
    c(event)
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/vegafusion_jupyter/widget.py", line 95, in _handle_message
    response_bytes = runtime.process_request_bytes(
  File "/Users/jonmmease/miniconda3/envs/try_solara/lib/python3.10/site-packages/vegafusion/runtime.py", line 128, in process_request_bytes
    return self.embedded_runtime.process_request_bytes(request)
[...]

PR #179, opened 2023-06-28
fix: hot reload was too aggresive

maartenbreddels:
Should fix
 * https://github.com/widgetti/solara/issues/177
 * https://github.com/widgetti/solara/issues/148

First step is to enable some tests that were flakey before.


solara/server/reload.py
```diff
@@ -152,25 +161,46 @@ def _on_change(self, name):
...
+        else:
+            reload = []
+            for name in sorted(sys.modules):
+                mod = sys.modules[name]
+                if name.startswith("solara.server"):
+                    continue  # this will break everything
+                if name in self.ignore_modules:
+                    continue  # nothing we imported from solara itself
+                if hasattr(mod, "__file__") and mod.__file__:
+                    if not mod.__file__.startswith(str(self.root_path)):
```
Lundez, line 179: if `mod.getattr("__file__", "MISSING").startswith(str(self.root_path))` would be cleaner IMO

maartenbreddels: thanks, I agree

maartenbreddels:
Thank you @Lundez for opening the issue and thinking along!

PR #184, opened 2023-06-29
docs: add a howto for debuggers

maartenbreddels:


solara/website/pages/docs/content/10-howto/51-debugging.md
```diff
@@ -0,0 +1,66 @@
...
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Solara: Launch",
+            "type": "python",
+            "request": "launch",
+            "program": "/Users/maartenbreddels/miniconda3/envs/dev/bin/solara",
```
Corvince, line 54: I just tested debugging and then saw that you have your program path here - this obviously needs to be adapted 
This should probably be documented, but otherwise it worked!

maartenbreddels: Nice to hear. Do you want to make a PR for this? You found it, so it would be fair you get the credits!

Issue #174, opened 2023-06-24
feat: Switch component


maartenbreddels:
Would be nice to wrap https://v2.vuetifyjs.com/en/components/switches/ using ipyvuetify, similar to checkbox https://github.com/widgetti/solara/blob/9cac7202072ab3fb3b316292501a1b880cc828de/solara/components/checkbox.py#L10 

See https://github.com/widgetti/solara/commit/8480176b66f43c31b3c384d48ec9f35287f31ffc for the original commit that added the checkbox.

maartenbreddels:
Closed by #185

Issue #131, opened 2023-05-31
test/ci issue: coverage slows down some tests


maartenbreddels:
After 172cdefeabd88f166d451873ea4582589a4cbb9b the test_memoize_hook fails more regularly. We've seen it fail before, but now it's almost 90%. Therefore we disabled coverage in CI for now.
If we want to enable it again, 172cdefeabd88f166d451873ea4582589a4cbb9b might give a clue.
This fails about 50% of the time on my osx laptop.
```
$ py.test --cov=solara tests/unit/cache_test.py
```

maartenbreddels:
ideally, we revert f84a001055c0b601492195b5536b62cde31be30d

PR #185, opened 2023-06-30
Feat: add switch widget

lp9052:
add switch widget

see #174


lp9052:
Do we have permission to rerun failed test?

maartenbreddels:
Thank you 🙏 

Apparently, you need to have write access for the repo to do that. I'm happy to restart it though

solara/components/switch.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@solara.value_component(bool)
+def Switch(
+    *,
+    label=None,
+    value: Union[bool, solara.Reactive[bool]] = True,
+    on_value: Callable[[bool], None] = None,
+    disabled=False,
+    style: str = None,
```
maartenbreddels, line 15: See bc81d7d93a608438674118afb0f27183de956afd for how we prefer the style argument

Please also add classes.

solara/components/switch.py
```diff
@@ -0,0 +1,55 @@
...
+
+     * `label`: The label to display next to the switch.
+     * `value`: The current value of the switch (True or False).
+     * `on_value`: A callback that is called when the switch is toggled.
+     * `disabled`: If True, the switch is disabled and cannot be used.
+     * `style`: A string of CSS styles to apply to the switch.
+    """
+    reactive_value = solara.use_reactive(value, on_value)
+    del value, on_value
+    children = []
```
maartenbreddels, line 52: See https://github.com/widgetti/solara/blob/6934ee5108ad2a05d10a50ae8b3ad16dcca22fb4/solara/components/button.py#L11 for how we can support any child in addition to a label argument

maartenbreddels (changes requested):
Just a few minor points that would make this PR even more perfect 👍

lp9052:
Thanks Maarten. 

I guess checkbox is one of the earlier piece that doesn't have all the new styles. All the enhancements make sense to me. I shall  update checkbox too. I'll start a new PR for checkbox.

maartenbreddels:
> I guess checkbox is one of the earlier piece that doesn't have all the new styles. All the enhancements make sense to me. I shall update checkbox too. I'll start a new PR for checkbox.

indeed, and that would be great!

> one small concern, we are passing `[]` as default parameter, it's mutable.

Yes, that's also something I usually frown upon... we are using that in many places because it also keeps the typing simpler so lets stick to it for now and if we change it, we'll do it throughout the whole codebase.

lp9052:
hmmm, the testing suit seems somewhat unstable. 

also I want to open up an issue about unit test doesn't all pass out of the box.

it requires some other packages as solar enterprise, etc.

maartenbreddels:
Hmm, it's been a while since that test failed. Before we saw that one fail more often, see https://github.com/widgetti/solara/issues/131 for some comments/clues.
The integration test also has 1 flakey test.

lp9052:
> Hmm, it's been a while since that test failed. Before we saw that one fail more often, see #131 for some comments/clues.
> The integration test also has 1 flakey test.

it's a quite interesting issue. I always like to see coverage info. And in my opinion 100% coverage is a must if we want to push this project further and accept more contributors

maartenbreddels:
I agree. A possible option might be to explicitly set `intrusive_cancel` in https://solara.dev/api/use_thread and only test that intrusive_cancel feature separately. 
Still, I find it really suspicious/odd that even after 10 seconds it does not finish the thread execution. I've stared at this failure many times and thought about it many times, but it might need a fresh pair of eyes.

maartenbreddels (approved):
💯

maartenbreddels:
There are 2 few missing I think
TODO:
 - [ ] Add to solara/website/pages/api/__init__.py
 - [ ] Add a gif 

I'm happy to pick this up if this is too much to ask. But this would also make this PR a nice template for new components.

lp9052:
> There are 2 few missing I think
> 
> TODO:
> 
>  - [ ] Add to solara/website/pages/api/__init__.py
> 
>  - [ ] Add a gif to solara/website/public/api/
> 
> 
> 
> I'm happy to pick this up if this is too much to ask. But this would also make this PR a nice template for new components.

Give me a day or two. I might run into problem for gif. But I'll let you know. Thanks.

lp9052:
Added these two pieces per suggestion.

lp9052:
@maartenbreddels

maartenbreddels:
Thank you! 🙏

Issue #149, opened 2023-06-09
FileBrowser doesn't follow conditional rendering


lp9052:
## Use Case:
Give a BASE_PATH, user can only go into subfolders/files, but not go up levels.

## Try 1:
```
base_path, set_path = solar.use_state(BASE_PATH)

def setter(path):
    if some_condition:
         set_path(path)

FileBrowser(directory=base_path, on_directory_change=setter)
```
This won't work, since _directory_ only controls starting point but is not used to control current folder

## Try 2:
```
base_path, set_path = solar.use_state(BASE_PATH)
 
if BASE_PATH in base_path:
    FileBrowser(directory=base_path, on_directory_change =set_path)
else:
    FileBrowser(directory= BASE_PATH, on_directory_change =set_path)
```

This also will not work, since even the condition is triggered correctly, but the Filebrowser widget itself won't rerender.

Is there anyway to control what are the folders that users can/cannot get into?

Thanks.

PR #188, opened 2023-06-30
enhancement: modify file bowser unit test to make it easier to understand

lp9052:
see  #149

### DIFF
--- a/tests/unit/file_browser_test.py
+++ b/tests/unit/file_browser_test.py
@@ -197,31 +197,26 @@ def test_file_browser_control_directory():
         return path.is_dir() and not path.name.startswith("_")
 
     BASE_PATH = HERE.parent.parent
-    directory = solara.reactive(BASE_PATH)
 
     @solara.component
     def Page():
-        def protect():
-            def check_base_path(value):
-                if not str(value).startswith(str(BASE_PATH)):
-                    directory.value = BASE_PATH
+        def set_directory(path: Path) -> None:
+            directory.value = path if str(path).startswith(str(BASE_PATH)) else BASE_PATH
 
-            return directory.subscribe(check_base_path)
-
-        solara.use_effect(protect, [])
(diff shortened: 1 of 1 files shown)


lp9052:
in my own project I use ```use_reactive``` with call back instead of using ```subscribe``` and ```use_effect```. Which I find might be a little more straightforward and easier to understand

lp9052:
oopsie, what happened to all these actions :)

lp9052:
Pydantic just updated to 2.0. Shall we limit the version to <2.0 for now?

tests/unit/file_browser_test.py
```diff
@@ -197,31 +197,26 @@ def directory_filter(path: Path) -> bool:
...
-    directory = solara.reactive(BASE_PATH)
 
     @solara.component
     def Page():
-        def protect():
-            def check_base_path(value):
-                if not str(value).startswith(str(BASE_PATH)):
-                    directory.value = BASE_PATH
+        def set_directory(path: Path) -> None:
+            directory.value = path if str(path).startswith(str(BASE_PATH)) else BASE_PATH
```
maartenbreddels, line 204:
```suggestion
            directory.value = path if str(path).startswith(str(BASE_PATH)) else BASE_PATH
```
Just to make you aware that this will always assign and will always need to do a comparison. Not that it matters because it is a test, but this might be slower in cases where compare is slow

maartenbreddels (approved):
LGTM, thank you!

tests/unit/file_browser_test.py
lp9052: Hmm, I never thought about that. But it would also hard to imagine a path comparison could be too expensive. Do you have an example in mind?

maartenbreddels: No, just wanted to make you aware that the behaviour has changed. Happy to merge this as is!

Issue #187, opened 2023-06-30
There seems to be a breaking change in pydantic 2.0


clausd:
Seems there's an issue with the brand new pydantic 2.0 release and solara - can't run even helloworld apps - hitting this

```

pydantic.errors.PydanticImportError: `BaseSettings` has been moved to the `pydantic-settings` package. See https://docs.pydantic.dev/2.0/migration/#basesettings-has-moved-to-pydantic-settings for more details.

For further information visit https://errors.pydantic.dev/2.0/u/import-error
```

clausd:
Downgrading Pydantic to 1.* fixes the problem

astrofrog:
Running into this too

maartenbreddels:
Finally managed to release 1.18.0 which fixes this!

PR #189, opened 2023-06-30
fix: remove pydantic

lp9052:
testing migrate to pydantic 2.0

where pydantic_settings is seperate package see #187

I quickly changed everything to pedantic_settings. But Solara depends on vaex which still error out.

I think for now we shall cap pedantic version. @maartenbreddels


maartenbreddels:
Ok, this is not as simple as I hoped, see https://github.com/pydantic/pydantic/discussions/6497 

Let's try to get of pydantic. We simply need an easy way to configure things, it shouldn't cause installation issues or breakage.

maartenbreddels:
Thanks for starting this. In the end, we had to remove Pydantic, otherwise we'd have to choose between v1 and v2 and we don't want to do that for our users.

lp9052:
Sorry for the delay, I was preparing my PhD application. It's unfortunate that we have to adopt this internally. Their decision to make a big change in a popular package is baffling. It's basically ineffective until everyone switches to pydantic2. I was dinged about our failing builds this morning, so I'm glad this was addressed. I truly appreciate it.

rht:
Thank you for shipping this!

PR #190, opened 2023-06-30
fix: cap pydantic < 2.0

lp9052:
due to pydantic breaking change.

see #187


maartenbreddels:
I agree, we first merge this, get CI green and see if we can get it working with 2 very soon, or do a release with a cap (I hate caps, so I prefer not to release with it :P)

maartenbreddels:
Thanks!

lp9052:
Ouch. Thanks for deduplicate it. Can't believe I made this mistake. Should we alphabetically sort the dependencies list. To avoid future issue like this.

maartenbreddels:
Np 🙂
I prefer as little changes as possible to make it easier to track changes, but if it bothers you a lot, a single sort pr is ok.
I should have sorted it from the beginning…

PR #199, opened 2023-07-06
Minor fix (practive -> practice)

tylere:


maartenbreddels:
Thank you Tyler! Surprised that codespell didn't pick that up 🤔

PR #205, opened 2023-07-11
Fix address warnings on kernel initialized dispatch

lp9052:
solara\server\patch.py will throw AttributionError when app is None. This mostly show up in unit test.


lp9052:
@maartenbreddels could you please take a look. It's kinda annoying and misleading when we run unit test.

lp9052:
@maartenbreddels

maartenbreddels:
I hit the same issue, and I generalized it a bit more, thank you for starting this!

PR #209, opened 2023-07-13
spelling

MajorTal:
spelling mistake in docs


maartenbreddels:
Thank you, we have codespell enabled in CI/pre-commit, but this was probably before that.

Issue #133, opened 2023-06-01
Quickstart tutorial and Solara Server fails with "no session cookie" on Safari


jfoster17:
Following the [Quickstart example](https://solara.dev/docs/quickstart) when I try to run the application in solara server as `solara run sol.py`, it does not work in Safari, although it does work in Chrome and Firefox. Safari reports

> ERROR:    no session cookie

at the terminal and

> [Error] WebSocket connection to 'ws://localhost:8765/jupyter/api/kernels/solara-id/channels?session_id=84fac001-b905-413b-b3d8-116e5c1d88cc' failed: There was a bad response from the server.

at the javascript console. 

This is with Safari 16.4.1. This is possibly nothing to do with Solara, and might be [this bug](https://bugs.webkit.org/show_bug.cgi?id=255524).

Running the application from within Jupyter Lab works just fine in all browsers I tested.

maartenbreddels:
Thank you Jonathan for the report, I'll look into it!

PR #234, opened 2023-08-06
fix dev cookie on Safari

qingant:
The behavior to set Secure=True when serving from localhost would cause failure to set cookie on Safari (tested on 17.0) and thus cause it hang on loading animation.

This could be an issue of Safari but pretty amount of developer who work on Mac would be prevented when they first try Solara on Mac if Safari was set as default browser.

### DIFF
--- a/solara/server/server.py
+++ b/solara/server/server.py
@@ -36,6 +36,7 @@ cache_memory = solara.cache.Memory(max_items=128)
 def get_jinja_env(app_name: str) -> jinja2.Environment:
     jinja_loader = jinja2.FileSystemLoader(
         [
+            os.getenv('SOLARA_TEMPLATES_DIR') or 'templates',
             app.apps["__default__"].directory.parent / "templates",
             str(directory / "templates"),
         ]
--- a/solara/server/starlette.py
+++ b/solara/server/starlette.py
@@ -163,7 +163,6 @@ async def kernel_connection(ws: starlette.websockets.WebSocket):
         user = None
 
     if not session_id:
-        logger.error("not work for Safari")
         logger.error("no session cookie")
         await ws.close()
         return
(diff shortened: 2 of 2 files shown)


maartenbreddels:
If this is really happening on Safarai 17 I think this is a browser bug (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value says localhost is ok) , I do not have this issue on Safari 16.5.2.

But, as the comment explains, this is needed for iframe support, so we cannot merge this as is. We'll need to think how we can work around this issue, maybe some opt out?

qingant:
> If this is really happening on Safarai 17 I think this is a browser bug (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value says localhost is ok) , I do not have this issue on Safari 16.5.2.
> 
> But, as the comment explains, this is needed for iframe support, so we cannot merge this as is. We'll need to think how we can work around this issue, maybe some opt out?

thanks. could you please point me to the direction where I could add the command options ? or , change to localStorage when Safari detected ? recent days, there is a couple of bugs around cookie in Safari (cookie manager for streamlit is also broken on Safari.)

qingant:
@maartenbreddels 

seems there is already an open issue: https://github.com/widgetti/solara/issues/133

It actually didn't work on my Safari 16.x version. I am guessing the reason that your browser works is because you already have cookie populated from previous run.

maartenbreddels:
You are correct, if i remove my cookie I get the same issue. This is pretty bad!

It seems we are doing a smart default thing (samesite=none and secure=true) to make more things work out of the box, but this default smart behaviour does not work on safari indeed.
I think we need to have something like
 1. A way to detect Safari (seems sth like `if 'Safari' in user_agent and 'Chrome' not in user_agent and 'Chromium' not in user_agent:`). If safari, do not do the smart default.
 2. manually turn on or off the samesite=none and secure=true settings. Turning it on can be useful if we are behind a proxy and fail to detect we are using https.

 I usually start adding these kinds of options in the solara.server.settings module. If they are common I also expose it via the command line (I don't think we need it this time).

What do you think of this?

PR #235, opened 2023-08-07
add support for index.html customization

qingant:
It's very common that we would want to customize the index.html especially the <head> section since we actually can't actually customize it from application layer.

This PR would enable the developer to customize the index.html by providing their own `index.html.j2` in `os.getenv('SOLARA_TEMPLATES_DIR') or 'templates'`. The framework might not want to allow user to do all the customizations since there seems a lot of vue initiating logic in the template, if developer rewrote that, it will simply not work. But I feel it's still beneficial to have this, since this is the only way to resolve a real problem.

For my self, I am also imaging to use Solara as a mechnisim to render Python side components and handle the interaction between frondend/backend and one day I might really want to build the UI system entirely with my own React UI library wrapper (Matine for example), further work needed to make it possible but in concept seems the need is just to define the interface between the template and solara render function (to render things under an element in the template).

### DIFF
--- a/solara/server/server.py
+++ b/solara/server/server.py
@@ -36,6 +36,7 @@ cache_memory = solara.cache.Memory(max_items=128)
 def get_jinja_env(app_name: str) -> jinja2.Environment:
     jinja_loader = jinja2.FileSystemLoader(
         [
+            os.getenv('SOLARA_TEMPLATES_DIR') or 'templates',
             app.apps["__default__"].directory.parent / "templates",
             str(directory / "templates"),
         ]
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Hi,

Love al; the ideas you have, shows that you understand Solara :) Indeed I can see a future with a different template and a different UI library (e.g. MaterialUI).

But as discussed on discord, the feature you implement in this PR is already supported by providing a file in `../templates` as we do with the solara website at https://github.com/widgetti/solara/tree/master/solara/website/templates

Currently, this is not documented because I'm not 100% sure yet if the current template is stable and if jinja is something we always want to use. Changing it after we document it would be a breaking change.

Feedback on this solara feature would be appreciated!

Also, knowing we already support this, do you think this PR still make sense, if so, why?

Regards,

Maarten

maartenbreddels:
I hope i've explained well enough that this is already supported right now, if you think there still is something missing from Solara, let us know and feel free to re-open this PR

PR #269, opened 2023-09-03
fix: web example ipyleaflet_advanced

hkayabilisim:
Problem:
For basemap variants such as Esri.WorlTopoMap or
Stamen.Watercolor, map url contains {variant} keyword. When you provide such urls directly to ipyleaflet.TileLayer.element, the map is not displayed properly.

Fix:
To resolve the issue, url must first be built
via map.build_url() so the {variant} keyword is evaluated correctly, then the built url is given to TileLayer.


maartenbreddels:
Thank you very much!

PR #274, opened 2023-09-05
fix: a hot reload could mix reactive variables from solara internal ones

maartenbreddels:
This happens with the user reactive variable, which got a key like 'None:0'. After a hot reload, the counter for this type was reset, and a different reactive variable got the same key. The value for the user reactive variable was then a dict with the user values.


maartenbreddels:
cc @mangecoeur I think you've been hit by this issue also

Issue #231, opened 2023-08-03
Allow mandatory=False for togglebuttonsSingle


mangecoeur:
It would be useful to be able to set whether a selection is mandatory or not in the solara ToggleButtonsSingle widget.

PR #277, opened 2023-09-06
feat: add attribute `mandatory` for ToggleButtons

iisakkirotko:
Fixes issue https://github.com/widgetti/solara/issues/231. Covers both `ToggleButtonsSingle` and `ToggleButtonsMultiple`.

### DIFF
--- a/solara/components/togglebuttons.py
+++ b/solara/components/togglebuttons.py
@@ -1,6 +1,6 @@
 from typing import Callable, Dict, List, Optional, TypeVar, Union, cast, overload
 
-import ipyvuetify as v  # type: ignore
+import ipyvuetify as v
 import reacton
 from typing_extensions import Literal
 
@@ -41,37 +41,7 @@ def ToggleButtonsSingle(
     value: Union[Optional[T], solara.Reactive[Optional[T]]] = None,
     values: List[T] = ...,
     children: List[reacton.core.Element] = ...,
-    on_value: Optional[Callable[[T], None]] = ...,
-    dense: bool = ...,
-    mandatory: Literal[False] = ...,
-    classes: List[str] = ...,
-    style: Union[str, Dict[str, str], None] = ...,
-) -> reacton.core.ValueElement[v.BtnToggle, T]:
-    ...
-
-
-@overload
-@solara.value_component(None)
-def ToggleButtonsSingle(
-    value: Union[T, solara.Reactive[T]],
-    values: List[T] = ...,
-    children: List[reacton.core.Element] = ...,
-    on_value: Optional[Callable[[T], None]] = ...,
-    dense: bool = ...,
-    mandatory: Literal[True] = ...,
-    classes: List[str] = ...,
-    style: Union[str, Dict[str, str], None] = ...,
-) -> reacton.core.ValueElement[v.BtnToggle, T]:
-    ...
-
-
-@overload
-@solara.value_component(None)
-def ToggleButtonsSingle(
-    value: Union[Optional[T], solara.Reactive[Optional[T]]] = None,
-    values: List[T] = ...,
-    children: List[reacton.core.Element] = ...,
-    on_value: Optional[Callable[[T], None]] = ...,
+    on_value: Optional[Callable[[Optional[T]], None]] = ...,
     dense: bool = ...,
     mandatory: Literal[False] = ...,
     classes: List[str] = ...,
@@ -146,13 +116,12 @@ def ToggleButtonsSingle(
     class_ = _combine_classes(classes)
     style_flat = solara.util._flatten_style(style)
     # TODO: make type safe
+    # typing is ignored below due to an issue with the typing; The combination of value being T and on_value being of type Callback[[Optional[T]], None] is
+    # not allowed to be passed to use_reactive. We also do not allow this by using our overloads, but this information seems lost at this point by
+    # the typechecker
     reactive_value = solara.use_reactive(value, on_value)  # type: ignore
     children = [solara.Button(label=str(value)) for value in values] + children
     values = values + [_get_button_value(button) for button in children]  # type: ignore
-    # When mandatory = True, index should not be None, but we are letting the front-end take care of setting index to 0 because of a bug
-    # (see https://github.com/widgetti/solara/issues/282)
-    # TODO: set index to 0 on python side (after #282 is resolved)
-    index, set_index = solara.use_state_or_update(values.index(reactive_value.value) if reactive_value.value is not None else None, key="index")
     # When mandatory = True, index should not be None, but we are letting the front-end take care of setting index to 0 because of a bug
     # (see https://github.com/widgetti/solara/issues/282)
     # TODO: set index to 0 on python side (after #282 is resolved)
@@ -162,13 +131,9 @@ def ToggleButtonsSingle(
         set_index(index)
         if mandatory:
             value = values[index]
-            reactive_value.set(value)
         else:
-            if index is not None:
-                value = values[index]
-                reactive_value.set(value)
-            else:
-                reactive_value.set(None)
+            value = values[index] if index is not None else None
+        reactive_value.set(value)
 
     return cast(
         reacton.core.ValueElement[v.BtnToggle, T],
@@ -176,43 +141,12 @@ def ToggleButtonsSingle(
     )
 
 
-@overload
-@overload
 @solara.value_component(None)
 def ToggleButtonsMultiple(
-    value: Union[List[T], solara.Reactive[List[T]]],
-    values: List[T] = ...,
-    children: List[reacton.core.Element] = ...,
-    on_value: Union[Callable[[Optional[List[T]]], None], None] = ...,
-    dense: bool = ...,
-    mandatory: Literal[True] = ...,
-    classes: List[str] = ...,
-    style: Union[str, Dict[str, str], None] = ...,
-) -> reacton.core.ValueElement[v.BtnToggle, List[T]]:
-    ...
-
-
-@overload
-@solara.value_component(None)
-def ToggleButtonsMultiple(
-    value: Union[Optional[List[T]], solara.Reactive[Optional[List[T]]]] = [],
-    values: List[T] = ...,
-    children: List[reacton.core.Element] = ...,
-    on_value: Union[Callable[[Optional[List[T]]], None], None] = ...,
-    dense: bool = ...,
-    mandatory: Literal[False] = ...,
-    classes: List[str] = ...,
-    style: Union[str, Dict[str, str], None] = ...,
-) -> reacton.core.ValueElement[v.BtnToggle, List[T]]:
-    ...
-
-
-@solara.value_component(None)
-def ToggleButtonsMultiple(
-    value: Union[Optional[List[T]], solara.Reactive[List[T]], solara.Reactive[Optional[List[T]]]] = [],
+    value: Union[List[T], solara.Reactive[List[T]]] = [],
     values: List[T] = [],
     children: List[reacton.core.Element] = [],
-    on_value: Union[Callable[[Optional[List[T]]], None], None] = None,
+    on_value: Union[Callable[[List[T]], None], None] = None,
     dense: bool = False,
     mandatory: bool = False,
     classes: List[str] = [],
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Nice!

I think now we can also be more strict on typing, similar to what we do with `InputFloat` https://github.com/widgetti/solara/blob/c017423dabb7dc50627c80c3de1988c8e34b79eb/solara/components/input.py#L131

Do you want to look into that? (e.g. when mandatory=True, None is not valid anymore)

iisakkirotko:
I'll look into also implementing that @maartenbreddels!

solara/components/togglebuttons.py
```diff
@@ -1,7 +1,8 @@
-from typing import Callable, Dict, List, Optional, TypeVar, Union, cast
+from typing import Callable, Dict, List, Optional, TypeVar, Union, cast, overload
 
-import ipyvuetify as v
+import ipyvuetify as v  # type: ignore
```
maartenbreddels, line 3: why is the type ignore needed?

solara/components/togglebuttons.py
```diff
@@ -86,26 +149,72 @@ def Page():
...
     values = values + [_get_button_value(button) for button in children]  # type: ignore
-    index, set_index = solara.use_state_or_update(values.index(reactive_value.value) if reactive_value.value is not None else 0, key="index")
+    # When mandatory = True, index should not be None, but we are letting the front-end take care of setting index to 0 because of a bug
+    # (see https://github.com/widgetti/solara/issues/282)
+    # TODO: set index to 0 on python side (after #282 is resolved)
+    index, set_index = solara.use_state_or_update(values.index(reactive_value.value) if reactive_value.value is not None else None, key="index")
+    # When mandatory = True, index should not be None, but we are letting the front-end take care of setting index to 0 because of a bug
+    # (see https://github.com/widgetti/solara/issues/282)
+    # TODO: set index to 0 on python side (after #282 is resolved)
+    index, set_index = solara.use_state_or_update(values.index(reactive_value.value) if reactive_value.value is not None else None, key="index")
```
maartenbreddels, line 159:
```suggestion
    # When mandatory = True, index should not be None, but we are letting the front-end take care of setting index to 0 because of a bug
    # (see https://github.com/widgetti/solara/issues/282)
    # TODO: set index to 0 on python side (after #282 is resolved)
    index, set_index = solara.use_state_or_update(values.index(reactive_value.value) if reactive_value.value is not None else None, key="index")
```
duplicate code

solara/components/togglebuttons.py
```diff
@@ -86,26 +149,72 @@ def Page():
...
-        reactive_value.set(value)
+        if mandatory:
+            value = values[index]
+            reactive_value.set(value)
+        else:
+            if index is not None:
+                value = values[index]
+                reactive_value.set(value)
+            else:
+                reactive_value.set(None)
```
maartenbreddels, line 171:
```suggestion
        if mandatory:
            value = values[index]
        else:
            value = values[index] if index is not None else None
        reactive_value.set(value)
```
I think I find this easier to read, because of less lines, and it's clear that the condition only affects how we go from values to value.

solara/components/togglebuttons.py
```diff
@@ -19,13 +20,74 @@ def _get_button_value(button: reacton.core.Element):
...
+    children: List[reacton.core.Element] = ...,
+    on_value: Optional[Callable[[T], None]] = ...,
+    dense: bool = ...,
+    mandatory: Literal[False] = ...,
+    classes: List[str] = ...,
+    style: Union[str, Dict[str, str], None] = ...,
+) -> reacton.core.ValueElement[v.BtnToggle, T]:
+    ...
+
+
```
maartenbreddels, line 82:
```suggestion
@overload
@solara.value_component(None)
def ToggleButtonsSingle(
    value: Union[T, solara.Reactive[T]],
    values: List[T] = ...,
    children: List[reacton.core.Element] = ...,
    on_value: Optional[Callable[[T], None]] = ...,
    dense: bool = ...,
    mandatory: Literal[True] = ...,
    classes: List[str] = ...,
    style: Union[str, Dict[str, str], None] = ...,
) -> reacton.core.ValueElement[v.BtnToggle, T]:
    ...


@overload
@solara.value_component(None)
def ToggleButtonsSingle(
    value: Union[Optional[T], solara.Reactive[Optional[T]]] = None,
    values: List[T] = ...,
    children: List[reacton.core.Element] = ...,
    on_value: Optional[Callable[[T], None]] = ...,
    dense: bool = ...,
    mandatory: Literal[False] = ...,
    classes: List[str] = ...,
    style: Union[str, Dict[str, str], None] = ...,
) -> reacton.core.ValueElement[v.BtnToggle, T]:
    ...


```

We only need two overloads, I think, for mandatory is True and False.

solara/components/togglebuttons.py
```diff
@@ -19,13 +20,74 @@ def _get_button_value(button: reacton.core.Element):
...
+@overload
+@solara.value_component(None)
+def ToggleButtonsSingle(
+    value: Union[Optional[T], solara.Reactive[Optional[T]]] = None,
+    values: List[T] = ...,
+    children: List[reacton.core.Element] = ...,
+    on_value: Optional[Callable[[T], None]] = ...,
```
maartenbreddels, line 44:
```suggestion
    on_value: Optional[Callable[[Optional[T]], None]] = ...,
```
I think this should be the typing given our on_index code.

solara/components/togglebuttons.py
```diff
@@ -86,26 +149,72 @@ def Page():
...
 
 
+@overload
+@overload
 @solara.value_component(None)
 def ToggleButtonsMultiple(
     value: Union[List[T], solara.Reactive[List[T]]],
+    values: List[T] = ...,
+    children: List[reacton.core.Element] = ...,
+    on_value: Union[Callable[[Optional[List[T]]], None], None] = ...,
```
maartenbreddels, line 186:
```suggestion
    on_value: Union[Callable[[List[T]], None], None] = ...,
```
I think for mandatory is True, it should be a list of T

solara/components/togglebuttons.py
```diff
@@ -86,26 +149,72 @@ def Page():
...
+    ...
+
+
+@overload
+@solara.value_component(None)
+def ToggleButtonsMultiple(
+    value: Union[Optional[List[T]], solara.Reactive[Optional[List[T]]]] = [],
+    values: List[T] = ...,
+    children: List[reacton.core.Element] = ...,
+    on_value: Union[Callable[[Optional[List[T]]], None], None] = ...,
```
maartenbreddels, line 201:
```suggestion
    on_value: Union[Callable[[List[T]], None], None] = ...,
```
My guess is with mandatory is False, the value is an empty list, not None, am I right?

solara/components/togglebuttons.py
```diff
@@ -86,26 +149,72 @@ def Page():
...
+    ...
+
+
+@solara.value_component(None)
+def ToggleButtonsMultiple(
+    value: Union[Optional[List[T]], solara.Reactive[List[T]], solara.Reactive[Optional[List[T]]]] = [],
     values: List[T] = [],
     children: List[reacton.core.Element] = [],
-    on_value: Callable[[List[T]], None] = None,
+    on_value: Union[Callable[[Optional[List[T]]], None], None] = None,
```
maartenbreddels, line 215: Should then match typing above.

solara/components/togglebuttons.py
```diff
@@ -133,15 +242,19 @@ def Page():
...
      * `on_value`: Callback to call when the value changes.
      * `dense`: Whether to use a dense (smaller) style.
+     * `mandatory`: Whether selecting at least one element is mandatory.
      * `style`: CSS style to apply to the top level element.
      * `classes`: List of CSS classes to be applied to the top level element.
     """
     class_ = _combine_classes(classes)
     style_flat = solara.util._flatten_style(style)
-    reactive_value = solara.use_reactive(value, on_value)
+    # Similar approach to select.py
```
maartenbreddels, line 251:
```suggestion
    # See comment in ToggleButtonsSingle regarding the typing issue
```

Is this what you meant?

solara/components/togglebuttons.py
```diff
@@ -133,15 +242,19 @@ def Page():
...
+    reactive_value = solara.use_reactive(value, on_value)  # type: ignore
     children = [solara.Button(label=str(value)) for value in values] + children
     allvalues = values + [_get_button_value(button) for button in children]
-    indices, set_indices = solara.use_state_or_update([allvalues.index(k) for k in reactive_value.value], key="index")
+    indices, set_indices = solara.use_state_or_update(
+        [allvalues.index(k) for k in reactive_value.value] if reactive_value.value is not None else None, key="index"
```
maartenbreddels, line 256:
```suggestion
        [allvalues.index(k) for k in reactive_value.value] if reactive_value.value is not None else [], key="index"
```
I think this is more consistent, instead of having both an empty list and None as valid values, we only use []

maartenbreddels (changes requested):
Great to see this green! A few comments about typing and some minor details.

solara/components/togglebuttons.py
iisakkirotko: Either works, but an empty list is more intuitive, I think.

solara/components/togglebuttons.py
iisakkirotko: I don't see a comment regarding the typing issue in ToggleButtonsSingle?

solara/components/togglebuttons.py
```diff
@@ -146,17 +116,15 @@ def Page():
     class_ = _combine_classes(classes)
     style_flat = solara.util._flatten_style(style)
     # TODO: make type safe
+    # typing is ignored below due to an issue with the typing of use_reactive; Typechecker expects a different type than is given. If expected type is
+    # given instead, a similar error shows up in typing of argument.
```
maartenbreddels, line 120:
```suggestion
    # typing is ignored below due to an issue with the typing; The combination of value being T and on_value being of type Callback[[Optional[T]], None] is not allowed to be passed to use_reactive. We also do not allow this by using our overloads, but this information seems lost at this point by the typechecker```

File: solara/components/togglebuttons.py [hunk 12]
```diff
@@ -77,35 +109,75 @@ def Page():
...
+    mandatory: Literal[False] = ...,
+    classes: List[str] = ...,
+    style: Union[str, Dict[str, str], None] = ...,
+) -> reacton.core.ValueElement[v.BtnToggle, List[T]]:
+    ...
+
+
+@solara.value_component(None)
+def ToggleButtonsMultiple(
+    value: Union[Optional[List[T]], solara.Reactive[List[T]], solara.Reactive[Optional[List[T]]]] = [],
```
maartenbreddels, line 175: I think there is nothing to change/overload here. I don't think we should do Optional[List[T]], and allow None, but only use an empty list to encode nothing is selected. The only things that should be constrained when mandatory=True is that the length of the list should be of length 1, which we cannot express in the typing system (it is a runtime thing).

solara/components/togglebuttons.py
```diff
@@ -133,15 +205,19 @@ def Page():
...
     style_flat = solara.util._flatten_style(style)
-    reactive_value = solara.use_reactive(value, on_value)
+    # See comment regarding typing issue in ToggleButtonsSingle
+    reactive_value = solara.use_reactive(value, on_value)  # type: ignore
     children = [solara.Button(label=str(value)) for value in values] + children
     allvalues = values + [_get_button_value(button) for button in children]
-    indices, set_indices = solara.use_state_or_update([allvalues.index(k) for k in reactive_value.value], key="index")
+    indices, set_indices = solara.use_state_or_update(
+        [allvalues.index(k) for k in reactive_value.value] if reactive_value.value is not None else [], key="index"
+    )
```
maartenbreddels, line 220: this means we can also remove this

maartenbreddels (changes requested):
The biggest change, I think, is that we should not allow None for the `Multiple` case, what do you think?

solara/components/togglebuttons.py
iisakkirotko: So just to be clear, remove all overloading from ToggleButtonsMultiple?

maartenbreddels: Yes, unless I am mistaken, the typing does not depend on the boolean argument `mandatory` anymore.

iisakkirotko: Yeah, that's correct when we don't allow for `value` to be `None`

iisakkirotko: Regardless, the code still works for `value=None`, but I don't think we want to do anything about that.

maartenbreddels:
Awesome, thank you!

PR #278, opened 2023-09-06
fix: "sample dataset"-button always disabled

iisakkirotko:


maartenbreddels:
Thank you @iisakkirotko!
The url fix is slightly unrelated, but let's merge this!

iisakkirotko:
@maartenbreddels indeed, we spent a while talking about that with @mariobuikhuizen, but thought that it probably isn't worth opening another PR for a one character change.

PR #286, opened 2023-09-12
feat(lab): new component ConfirmationDialog

hansnowak:
First, very rough, version of `ConfirmationDialog`.


maartenbreddels:
Great to see this Hans. For code quality checks I recommend you follow the pre-commit steps as suggested in https://solara.dev/docs/development . This will run some linters when you commit for a faster feedback loop.

solara/lab/components/confirmation_dialog.py
```diff
@@ -0,0 +1,49 @@
...
+        on_close()  # possible additional actions when closing
+        is_open_rv.set(False)
+
+    if is_open_rv.value:
+        with v.Dialog(
+            #children=children,
+            v_model=is_open_rv.value,
+            on_v_model=is_open_rv.set,
+            permanent=True,
+            max_width=500,
```
maartenbreddels, line 32: @mariobuikhuizen any ideas on a good default width? If we do something like this (e.g. the default is not good enough) we should expose it as an argument.

solara/lab/components/confirmation_dialog.py
```diff
@@ -0,0 +1,49 @@
...
+                    solara.Markdown(content)
+                else:
+                    solara.display(content)
+                if children:
+                    solara.display(*children)
+                with solara.CardActions():
+                    if isinstance(ok, str):
+                        solara.Button(label=ok, on_click=perform_action)
+                    else:
+                        solara.display(ok)
```
maartenbreddels, line 45: I see an issue here is that a user provided button will not trigger perform_action. It would be nice to expose this during the tests. It might be that if we need to use `use_event` for this that we hit this bug: https://github.com/widgetti/reacton/issues/14

maartenbreddels (commented):
Hi Hans, I did a first pass. Generally, it looks good 👍 , I'm a bit uncertain about the naming (especially is_open_rv)

solara/lab/components/confirmation_dialog.py
mariobuikhuizen: yeah, I think we should expose `max_width`, as one would want to tailor it to the content. I think 500px is a good default.

solara/lab/components/confirmation_dialog.py
```diff
@@ -0,0 +1,49 @@
...
+    def close():
+        on_close()  # possible additional actions when closing
+        is_open_rv.set(False)
+
+    if is_open_rv.value:
+        with v.Dialog(
+            #children=children,
+            v_model=is_open_rv.value,
+            on_v_model=is_open_rv.set,
+            permanent=True,
```
mariobuikhuizen, line 31: I don't think `permanent` is argument of `Dialog`

iisakkirotko: Should be `persistent`?

solara/lab/components/confirmation_dialog.py
```diff
@@ -0,0 +1,49 @@
+from typing import cast, Callable, List, Union, Optional
+
+import solara
+import reacton.ipyvuetify as v
+
+
+@solara.component
+def ConfirmationDialog(
+    is_open_rv: solara.Reactive[bool],
```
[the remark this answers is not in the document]
hansnowak: Hm, `value` seems a bit too vague in my opinion... that makes sense for a switch etc, but maybe less so for a dialog... what is the "value" of a dialog?? It might be better to use something with "open". `is_open` seems to be more in line with what is usually used in Python, indicating a boolean. 

I'm not sure what the callback is for? Is it called when `is_open` (or whatever we end up calling it) changes value?

solara/lab/components/confirmation_dialog.py
```diff
@@ -0,0 +1,49 @@
...
+    action: Callable[[], None],
```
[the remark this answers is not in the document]
hansnowak: OK! :) The other one should probably be `on_cancel` then...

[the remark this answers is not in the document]
hansnowak: Or does `on_cancel` or `on_close` (as it is currently called) become redundant?

solara/lab/components/confirmation_dialog.py
hansnowak: Yeah. I don't know where I got "permanent" from. I changed it; should it be exposed though?

solara/lab/components/confirmation_dialog.py
maartenbreddels: Yes, as I mentioned in our call, if you write a test we can find a solution to this later on, I have some ideas about it.

solara/lab/components/confirmation_dialog.py
maartenbreddels: Yes, I think so, especially if we change the default.

solara/lab/components/confirmation_dialog.py
maartenbreddels: In theory, we should be able to use `solara.v.use_event`, but it has issues due to https://github.com/widgetti/reacton/issues/14

I think we can do a different approach, which is to inspect ok.kwargs["on_click"] and replace it by a new dict where on_click points to our perform_action, and the perform_action also calls the previous `on_click` set by the user.

hansnowak: So, if I understand this correctly, if you pass a custom `Button` named `ok`, then `ok.kwargs['on_click']` is only present if you passed an `on_click` event to your button. Should we allow for a case where a user passes in a custom button that has *no* `on_click`? Perhaps because they just want to execute `on_ok` as intended.

Related: what do we do in the case that both `on_ok` and a custom button's `on_click` are set?

hansnowak: For context, I currently have this:

```python
    with solara.CardActions():
        if isinstance(ok, str):
            solara.Button(label=ok, on_click=perform_action)
        else:
            solara.display(ok)
            if ok.kwargs.get('on_click'):
                ok.kwargs['on_click'] = perform_action
```

...which passes the test. I think there is no point in checking `ok.kwargs` if `ok` is a string?

maartenbreddels: Yes, I think we should support no on_click as well. I suggest we also don't modify the dict, but create a new one.

hansnowak: > In theory, we should be able to use `solara.v.use_event`, but it has issues due to [widgetti/reacton#14](https://github.com/widgetti/reacton/issues/14)
> 
> I think we can do a different approach, which is to inspect ok.kwargs["on_click"] and replace it by a new dict where on_click points to our perform_action, and the perform_action also calls the previous `on_click` set by the user.

I should have read this more carefully, you're literally describing what to do in a situation with both `on_ok` and `on_click` being present.

maartenbreddels:
This is released in v1.20.0
Thank you Hans!

icduck:
This is a nice component. Before this, I used v.Dialog which is good too. this ConfirmationDialog would be great if it supports content_class just like v.Dialog and draggable feature. Currently, I adapted this https://github.com/vuetifyjs/vuetify/issues/4058 for draggable dialog in Solara. Love to see it as built-in feature for dialog, just like griddraggable.

Issue #287, opened 2023-09-12
FigureAltair does not render in jupyter/colab


rlskoeser:
I'm working one some some code generating altair charts and they display just fine when I use the local solara runserver, but when I try to render them in jupyter and colab, nothing displays. 

I don't see any python errors to help diagnose the problem; in the browser javascript console for Colab, I see these javascript errors:
```
ReferenceError: requirejs is not defined
...
TypeError: this.do_plot_debounced is not a function
...
```

I installed altair 5.1.1 in Colab to match the version in my local python environment.  

I've done some googling for these errors but not found anything that seemed directly applicable.

Please advise on how to resolve or how else I can troubleshoot. Thank you!

tpike3:
I believe I am having a similar issue as @rlskoeser I can get the Altair example to render wrapping the example in code in display --

`display(alt.Chart(source, title="Daily Max Temperatures (C) in Seattle, WA").mark_rect().encode(
    alt.X("date(date):O").title("Day").axis(format="%e", labelAngle=0),
    alt.Y("month(date):O").title("Month"),
    alt.Color("max(temp_max)").title(None),
    tooltip=[
        alt.Tooltip("monthdate(date)", title="Date"),
        alt.Tooltip("max(temp_max)", title="Max Temp"),
    ],
).configure_view(
    step=13,
    strokeWidth=0
).configure_axis(
    domain=False
))`


 but the chart doesn't dynamically update like with the .py files rendered in the browser. This is the same for the `ipyleaflet advanced` where the map is dynamic in that I can move the marker but the location does not update and then if I click the reset or zoom marker the other widgets will update (i.e. zoom level) but the map does not.

maartenbreddels:
Hi Rebecca,

thanks for opening the issue. Are you using Jupyter Lab of notebook, and which version are you using.

@tpike3 I do not think the ipyleaflet issue is related, could you open a new issue for that and precisely describe how I can reproduce it (do you use solara server/jupyter lab/notebook, and which versions)

Regards,

Maarten

rlskoeser:
Hi @maartenbreddels, thanks for responding. 

I was initially using Colab. I don't know if there's a way to see anything about version information there, do you know? When it didn't work there, I tested with a Jupyter notebook running through Jupyter Lab version 4.0.1.  I could try upgrading that if you think it would help, but I do want to be able use this in Colab.

rlskoeser:
[...]

PR #288, opened 2023-09-13
fix: make altair work in Jupyter notebook and Jupyter Lab

maartenbreddels:
1. notebook: wrong mimetype (see 1st commit)
2. lab: requirejs was not loaded (see 2nd commit)

Fixes the non-colab part of #287


maartenbreddels:
failures are unrelated

Issue #281, opened 2023-09-07
Autoreload for subpackages


Jhsmit:
When you have an application with the following structure:

`my_application/app`  (multipage solara app, directory with `__init__.py` which has `Page` component
`my_application/components` (module with solara components used in solara app)

Then when running as `solara run my_application.app`, and making changes in `components`,  autoreload is triggered, but the change is not seen in the reloaded application.

The desired behavior is that all changes in the complete package are reloaded, not the subpackage only. 

Workaround for testing/development is to create a file higher in the directory hierarchy and run from there.

Issue #242, opened 2023-08-11
New instance of a Reactive object after hot-reload


mariobuikhuizen:
After a hot reload, top level variables with Reactive objects get new instances. This is a problem when a variable was directly imported, which would still reference the old instance.

PR #293, opened 2023-09-15
fix: subpackage changes not autoreloaded

iisakkirotko:
Fixes issue #281.

### DIFF
--- a/solara/server/app.py
+++ b/solara/server/app.py
@@ -82,8 +82,9 @@ class AppScript:
         if self.type == AppType.MODULE:
             package_name = self.name.split(".")[0]
             mod = importlib.import_module(package_name)
-            package_root_path = Path(mod.__file__).parent
-            reload.reloader.root_path = package_root_path
+            if mod.__file__ is not None:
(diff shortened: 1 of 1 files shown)


solara/server/app.py
```diff
@@ -79,6 +79,11 @@ def __init__(self, name, default_app_name="Page"):
 
         self._first_execute_app = app
         reload.reloader.root_path = self.directory
+        if self.type == AppType.MODULE:
+            package_name = self.name.split(".")[0]
+            mod = importlib.import_module(package_name)
+            package_root_path = Path(mod.__file__).parent
+            reload.reloader.root_path = package_root_path
```
maartenbreddels, line 86: I think this needs to be conditional (looking at the mypy error) when mod.__file__ is None, we just skip this I think.

iisakkirotko: Made the change. Shouldn't worry too much about the ~30 errors from Python 3.9?

maartenbreddels:
Also fixes #242

maartenbreddels:
Thanks 👍

Issue #296, opened 2023-09-20
Cannot switch to dark mode


Jhsmit:
running an app with the option `theme-variant` should enable dark theme, but this does not work for me.
Clean install on win11, reproduced by @mariobuikhuizen 

```
$ solara run --theme-variant dark script.py
```

Issue #156 suggests that it did work in previous versions

Ranger11Danger:
@Jhsmit  i'm having this problem too and found a work around until its fixed
if you goto where your python library is intalled `/home/<username>/.local/lib/<python version>/site-packages/solara/server`
there is a file `settings.py` on line 33 you can change `variant: ThemeVariant = ThemeVariant.light` -> `variant: ThemeVariant = ThemeVariant.dark` to make it default to dark

maartenbreddels:
This should be fixed in v1.20.0

PR #303, opened 2023-09-27
fix: pass settings to the jinja template. 

maartenbreddels:
Because we did not implement a setter for the settings, command
line arguments were not passed to the jinja template.
Values were set, but .dict() used the _values dict which was
not updated by the setter.

Note that environment variables did work, so although
SOLARA_THEME_VARIANT=dark HOST=localhost solara run ...
Did work, the following did not:
HOST=localhost solara run ... --theme-variant=dark

### DIFF
--- a/solara/minisettings.py
+++ b/solara/minisettings.py
@@ -56,6 +56,9 @@ class _Field:
             return self
         return instance._values[self.name]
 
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Related to https://github.com/widgetti/solara/issues/296

PR #305, opened 2023-09-27
ci: fix lint failure of mypy due to modern typed libraries

maartenbreddels:
Introduced in traitlets 5.10 due to
 * https://github.com/ipython/traitlets/pull/868
 * https://github.com/ipython/traitlets/pull/818

The correct way would be to fix the errors, but we don't want CI to fail randomly. This also supports our idea of having lockfiles for CI.

### DIFF
--- a/.github/workflows/codequality.yaml
+++ b/.github/workflows/codequality.yaml
@@ -23,7 +23,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install ".[dev]" mypy==0.991 black==22.12.0 codespell==2.2.4 "click<8.1.4"
+          pip install ".[dev]" mypy==0.991 black==22.12.0 codespell==2.2.4 "click<8.1.4" "traitlets<5.10.0" "matplotlib<3.8.0"
           mypy --install-types --non-interactive solara
       - name: Run codespell
         run: codespell


maartenbreddels:
Also matplotlib 3.8 seems to introduce typing which makes mypy linting fail.

EwoutH:
This should be tagged as technical debt and, in my opinion, prioritized.

.github/workflows/codequality.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          pip install ".[dev]" mypy==0.991 black==22.12.0 codespell==2.2.4 "click<8.1.4"
+          pip install ".[dev]" mypy==0.991 black==22.12.0 codespell==2.2.4 "click<8.1.4" "traitlets<5.10.0" "matplotlib<3.8.0"
```
EwoutH, line 26: Versions are currently pinned in the GitHub workflow, but not in the [`pyproject.toml`](https://github.com/widgetti/solara/blob/master/pyproject.toml) (optional) dependencies. Doesn't that mean users can install dependency versions that are not supported or tested by solara?

As far as I know, best practice is to define dependencies in `pyproject.toml` and use those in testing/CI. If you have dependencies only used in testing or dev, add them to an optional dependency list.

maartenbreddels: I agree that we should fix this soon, but it's not as high a priority as a runtime failure.

However, I disagree we should do version pinning in solara since that can make environments unsolvable. Pinning should only be done by applications (or application-like libraries, or when you know for sure there is a failure.

EwoutH: Thanks for getting back. I agree that pinning exact versions is bad practice (for exactly the reasons you said), 

Also I can't imagine black and mypy should be required dependencies, those should be optional dependencies (in `.[dev]` or `.[ci]` for example), and then regular users won't install them. Thus limiting the version range doesn't affect them, since it's purely for internal use.

PR #325, opened 2023-10-11
fix: strange menu behaviour with some child components

iisakkirotko:
Because `on_show_menu` hadn't been set, Python and Vue weren't in sync about the status of a menu being open.

### DIFF
--- a/solara/lab/components/menu.py
+++ b/solara/lab/components/menu.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 
 import solara
 from solara.components.component_vue import component_vue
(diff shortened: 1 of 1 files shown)


solara/lab/components/menu.py
```diff
@@ -123,13 +142,23 @@ def Page():
     ## Arguments
 
     * activator: Clicking on this element will open the menu. Accepts either a `solara.Element`, or a list of elements.
+    * open: Controls and communicates the state of the menu. If True, the menu is open. If False, the menu is closed.
     * children: List of Elements to be contained in the menu
     * style: CSS style to apply. Applied directly onto the `v-menu` component.
+    * close_on_content_click: If True, clicking on the menu contents will close the menu. If False, the menu will stay open.
```
maartenbreddels, line 148: This argument is missing, I see it in #326 but you probably wanted it here.

maartenbreddels (changes requested):
Nice work, just a small error I think. What happens with the ClickMenu and ContextMenu when you manually set open (or open.value) to True?

solara/lab/components/menu.py
iisakkirotko: My thought was that `close_on_content_click` was something to allow `InputDate` to work, and therefore shouldn't be here but only in #326. Let me know your thoughts.

maartenbreddels: Ah yes, true, so we should only move the docstring there

maartenbreddels:
Perfect 👍

maartenbreddels (approved):

PR #329, opened 2023-10-19
Fixed the logic that is in the first example of the quick start guide

draperjames:
The current example of the first app in the quick start guide has logic that does not yield correct results: https://solara.dev/docs/quickstart

These two lines fix that issue.


maartenbreddels (commented):
Excellent, I love your attention to detail, I just left a question for the 3rd line you changed.

PR #338, opened 2023-10-24
feat: Improved and generalized `_use_input_numeric`

iisakkirotko:
The hook is now much more legible, and lends itself to more generic use cases.

### DIFF
--- a/solara/components/input.py
+++ b/solara/components/input.py
@@ -1,4 +1,3 @@
-import ast
 from typing import Any, Callable, Optional, TypeVar, Union, cast, overload
 
 import ipyvue
@@ -197,15 +196,28 @@ def InputFloat(
     * `clearable`: Whether the input can be cleared.
 
     """
+
+    def str_to_float(value: Optional[str]):
+        if value is not None and value != "":
+            try:
+                value = value.replace(",", ".")
+                return float(value)
+            except ValueError:
+                raise ValueError("Value must be a number")
+        else:
+            if optional:
+                return None
+            else:
+                raise ValueError("Value cannot be empty")
+
     return _InputNumeric(
-        float,
+        str_to_float,
         label=label,
         value=value,
         on_value=on_value,
         disabled=disabled,
         continuous_update=continuous_update,
         clearable=clearable,
-        optional=optional,
     )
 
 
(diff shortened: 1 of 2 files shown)


solara/components/input.py
```diff
@@ -197,15 +196,28 @@ def Page():
     * `clearable`: Whether the input can be cleared.
 
     """
+
+    def str_to_float(value: Optional[str]):
+        if value is not None and value != "":
```
maartenbreddels, line 201: I think this ideas easier
```suggestion
        if value:
```

solara/components/input.py
```diff
@@ -276,113 +288,70 @@ def Page():
     * `continuous_update`: Whether to call the `on_value` callback on every change or only when the input loses focus or the enter key is pressed.
     * `clearable`: Whether the input can be cleared.
     """
+
+    def str_to_int(value: Optional[str]):
+        if value is not None and value != "":
```
maartenbreddels, line 293:
```suggestion
        if value:
```

tests/unit/input_test.py
```diff
@@ -106,34 +106,30 @@ def update_value(value: int):
     assert on_value.call_count == 0
     assert on_v_model.call_count == 1
     input.widget.fire_event("blur")
-    assert on_value.call_count == 1
-    assert on_value.call_args[0][0] == 1000
-    assert not input.widget.error
-    assert input.widget.label == "label"
+    assert on_value.call_count == 0
+    # assert on_value.call_args[0][0] == 1000
```
maartenbreddels, line 110:
```suggestion
```

tests/unit/input_test.py
```diff
@@ -106,34 +106,30 @@ def update_value(value: int):
...
+    assert not input.widget.error
     input.widget.fire_event("blur")
     # no change
-    assert on_value.call_count == 2
-    assert on_v_model.call_count == 5
-    assert not input.widget.error
-    assert input.widget.label == "label"
-    assert input.widget.v_model == "1"
+    assert on_value.call_count == 1
+    # assert on_v_model.call_count == 5
```
maartenbreddels, line 129:
```suggestion
```

tests/unit/input_test.py
```diff
@@ -106,34 +106,30 @@ def update_value(value: int):
...
+    assert input.widget.error
+    assert input.widget.label == "label (Value must be an integer)"
+    # assert input.widget.v_model == "1"
```
maartenbreddels, line 132:
```suggestion
    assert input.widget.v_model == "1.1"
```

solara/components/input.py
```diff
@@ -276,113 +288,70 @@ def Page():
...
-                set_error(True)
-            return
-        try:
-            numeric_value = str_to_numeric(parse(value))
-        except Exception:
-            set_error(True)
-        else:
-            set_error(False)
-            reactive_value.set(numeric_value)
+    solara.use_effect(hook, [reactive_value.value])
```
maartenbreddels, line 343:
```suggestion
    solara.use_effect(sync_back_input_value, [reactive_value])
```
This this only re-attach when the reactive_value object changes (very uncommon edge case that can happen when the parent component sometimes passes a reactive var, and sometime a normal var). It doesn't need to re-attach when the value changes.

solara/components/input.py
```diff
@@ -276,113 +288,70 @@ def Page():
...
     on_value: Union[None, Callable[[Optional[T]], None], Callable[[T], None]] = None,
-    optional: bool = False,
 ):
-    # TODO: make this more type safe
-    reactive_value = solara.use_reactive(value, on_value)  # type: ignore
-    del value, on_value
+    reactive_value = solara.use_reactive(input_value, on_value)  # type: ignore
+    del input_value, on_value
+    string_value, set_string_value = solara.use_state(stringify(reactive_value.value) if reactive_value.value is not None else None)
+    # Workaround to hook() getting a stale string_value
```
maartenbreddels, line 324:
```suggestion
    # Use a ref to make sure sync_back_input_value() does not get a stale string_value
```
it's not a workaround I would say, but it's more like a common pattern in react

solara/components/input.py
```diff
@@ -276,113 +288,70 @@ def Page():
...
-            except Exception:
-                return internal_value
-            else:
-                if numerical_proper_type != numerical:
-                    return str(numerical_proper_type)
-                else:
-                    return internal_value
-        else:
-            return internal_value
+    def hook():
```
maartenbreddels, line 335:
```suggestion
    def sync_back_input_value():
```

solara/components/input.py
```diff
@@ -276,113 +288,70 @@ def Page():
...
+        def on_external_value_change(new_value: Optional[T]):
+            new_string_value = stringify(new_value)
+            if new_value != parse(string_value_ref.current):
```
maartenbreddels, line 338: Lets add a test to cover a possible problem here:
Enter invalid input, e.g. "1a", not change the reactive variable we passed (e.g. we choose to modify a slider that is linked to the same reactive var). Now this triggers an exception.

maartenbreddels (changes requested):
Awesome, looks so much cleaner

solara/components/input.py
iisakkirotko: Fixed and tested for by new commit.

maartenbreddels:
Nice, thank you!

PR #345, opened 2023-10-26
Fix: added width option and removed additional `<div>` around activator

iisakkirotko:


maartenbreddels:
seems like CI is broken, also https://github.com/widgetti/solara/pull/350 is failing

maartenbreddels (changes requested):
Really happy with these changes, just some small minor comments.

solara/lab/components/menu.py
```diff
@@ -154,6 +156,8 @@ def Page():
     * on_open_value: Function to call when the menu is opened or closed.
     * children: List of Elements to be contained in the menu
     * style: CSS style to apply. Applied directly onto the `v-menu` component.
+    * use_activator_width: If True, the menu will have a minimum width equal to the activator element.
+    If False, the menu width will be determined by the content.
```
maartenbreddels, line 160:
```suggestion
      If False, the menu width will be determined by the content.
```
requires indentation to be under the same bullet afaik

PR #346, opened 2023-10-26
CI: Test against Python 3.10

EwoutH:
Two bits of maintenance on the CI:
- In GitHub Actions, test against Python 3.10
- Update ipywidgets to the latest minor versions (7.8 and 8.1 respectively)

Testing against a more modern Python version ensures it not only works on that version, but also gives warnings about potential incompatibilities with future Python versions. Since some packages are already dropping support for Python 3.9 and older (see [SPEC 0](https://scientific-python.org/specs/spec-0000/)), testing at least against Python 3.10 is recommended.

Testing against Python 3.11 and Python 3.12 is currently blocked by Vaex not supporting those versions, see https://github.com/vaexio/vaex/pull/2331. I don't know exactly what Vaex is used for in Solara, but it could be worth considering removing it as a dependency.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -25,14 +25,13 @@ jobs:
         # just ubuntu and windows give enough confidence
         # osx should work fine (and we test that locally often)
         os: [ubuntu, windows]
-        # just 1 version, it's heavy
-        python-version: [3.8]
-        ipywidgets: ["7.7", "8.0"]
         include:
-          - ipywidgets: "7.7"
-            voila: "0.3.0"
-          - ipywidgets: "8.0"
-            voila: "0.4.0"
+          - ipywidgets: "7.8"
+            voila: "0.3.7"
+            python-version: "3.10"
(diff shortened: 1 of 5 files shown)


.github/workflows/integration.yml
```diff
@@ -25,14 +25,13 @@ jobs:
...
-          - ipywidgets: "7.7"
-            voila: "0.3.0"
-          - ipywidgets: "8.0"
-            voila: "0.4.0"
+          - ipywidgets: "7.8"
+            voila: "0.3.7"
+            python-version: "3.10"
+          - ipywidgets: "8.1"
+            voila: "0.5.4"
+            python-version: "3.10"
```
maartenbreddels, line 34: Note that we use a ~ for the installation, so this change is not needed.

maartenbreddels: Ah, I see that you have changed the matrix from 2x2 to 1x2. I think ipywidsgets should be kept top level for the matrix to work.

PR #347, opened 2023-10-26
CI: Run workflows once a week

EwoutH:
Adds a weekly scheduled run to all CI workflows which run on push or PR, which now run each Monday at 06:00 UTC.

A periodically scheduled run can detect errors and warnings appearing in changes in upstream dependencies or the build environment. By having a scheduled run they can be traced back easier to a dependency or environment change, then if they would pop up in a random PR.

### DIFF
--- a/.github/workflows/codequality.yaml
+++ b/.github/workflows/codequality.yaml
@@ -6,6 +6,8 @@ on:
       - master
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'
 
 jobs:
   code-quality:
--- a/.github/workflows/codequality.yaml
+++ b/.github/workflows/codequality.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
   workflow_dispatch:
   schedule:
-    - cron: '0 6 * * 1'
+    - cron: '0 6 * * 1,4'  # at 06:00 UTC on Monday and Thursday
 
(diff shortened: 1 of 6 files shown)


maartenbreddels:
Thank you!
Note that we have bigger plans along the lines or https://github.com/widgetti/ipyaggrid/issues/44#issuecomment-1689710768 but I think this is indeed a good improvement.
Although I think once a week is maybe not often enough. What about every day? Or twice a week?

EwoutH:
Thanks for getting back! Since these workflows also run on every push to the main branch and on every PR, I think two additional runs a weeks should be more than enough. It’s more to have some recent reference run in periods of inactivity.

maartenbreddels:
Could you add comments to the cron lines? I've never met someone in my life (except chatgpt) who knows by heart what it means. :)

EwoutH:
Yes, will do!

I always use https://crontab.guru/ by the way, I find it really useful.

maartenbreddels:
Dankjewel Ewout!

EwoutH:
Graag gedaan!

EwoutH:
The first scheduled runs where queued and performed [as expected](https://github.com/widgetti/solara/actions?query=event%3Aschedule).

Screenshot_203

maartenbreddels:
Love it ❤️ ! Thank you

PR #348, opened 2023-10-26
pytest config: remove `--disable-warnings`

EwoutH:
Remove the `--disable-warnings` argument from the pytest configuration in `pyproject.toml`.

This will re-enable warmings in the unit and integrations tests, allowing deprecations and future unsupported behaviour to be noticed before they turn into errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,7 @@ profile = "black"
 
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Yeah, i like that. We probably have to allow for some warnings, but I agree with the general point 👍

EwoutH:
Correct, is only shows the warnings instead of suppressing them.

Now that the warnings are shown, they should be resolved. I opened an issue to track those: https://github.com/widgetti/solara/issues/360

PR #352, opened 2023-10-27
refactor: replace --dev with --auto-restart and have a --production mode

maartenbreddels:
Before, we had a --dev mode that would auto-restart the server when
files changed. This means that auto restart was connected to dev mode.
Now, by default solara is in development mode, and the --auto-restart
option is separated.
This means that we now have a special --production mode that should
be used in production settings. It will serve users with minimized
JS and CSS and disable project and .vue reloading.

This means --reload (a bad and confusing name) and --dev are now
deprecated, but will behave the same as --auto-reload/-a.

### DIFF
--- a/solara/__main__.py
+++ b/solara/__main__.py
@@ -123,40 +123,34 @@ def cli():
     default=settings.main.host,
     help="Host to listen on. Defaults to the $HOST environment or $SOLARA_HOST when available or localhost when not given.",
 )
-@click.option(
-    "--dev/--no-dev",
-    default=False,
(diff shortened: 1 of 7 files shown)


maartenbreddels:
* master
  * **PR #352** <a href="https://app.graphite.dev/github/pr/widgetti/solara/352" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/352?utm_source=stack-comment).

PR #372, opened 2023-11-10
docs: added documentation for hooks

iisakkirotko:
Added documentation for `use_effect`, `use_memo`, and improved the documentation of `use_thread`

### DIFF
--- a/solara/website/pages/api/use_memo.md
+++ b/solara/website/pages/api/use_memo.md
@@ -9,6 +9,8 @@ def use_memo(
     ...
 ```
 
-`use_memo` stores ([memoize](/api/memoize)) the function return on first render, and then excludes it from being re-executed, except when one of the `dependencies` changes. `dependencies` can take the value `None`, in which case dependencies are automatically obtained from nonlocal variables. If an empty list is passed as `dependencies` instead, the function is only executed once over the entire lifetime of the component.
+`use_memo` stores ([memoize](https://en.wikipedia.org/wiki/Memoization)) the function return on first render, and then excludes it from being re-executed, except when one of the `dependencies` changes. `dependencies` can take the value `None`, in which case dependencies are automatically obtained from nonlocal variables. If an empty list is passed as `dependencies` instead, the function is only executed once over the entire lifetime of the component.
+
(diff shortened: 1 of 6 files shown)


iisakkirotko:
* master
  * **PR #372** <a href="https://app.graphite.dev/github/pr/widgetti/solara/372?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/372?utm_source=stack-comment).

solara/website/pages/api/use_memo.md
```diff
@@ -0,0 +1,14 @@
...
+```python
+def use_memo(
+    f: Any,
+    dependencies: Any | None = None,
+    debug_name: str = None
+) -> Any:
+    ...
+```
+
+`use_memo` stores ([memoize](/api/memoize)) the function return on first render, and then excludes it from being re-executed, except when one of the `dependencies` changes. `dependencies` can take the value `None`, in which case dependencies are automatically obtained from nonlocal variables. If an empty list is passed as `dependencies` instead, the function is only executed once over the entire lifetime of the component.
```
maartenbreddels, line 12:
```suggestion
`use_memo` stores ([memoize](https://en.wikipedia.org/wiki/Memoization)) the function return on first render, and then excludes it from being re-executed, except when one of the `dependencies` changes. `dependencies` can take the value `None`, in which case dependencies are automatically obtained from nonlocal variables. If an empty list is passed as `dependencies` instead, the function is only executed once over the entire lifetime of the component.

Not to be confused with [memorize](https://solara.dev/api/memoize) which can cache multiple return values and which can be used outside of component.
```

maartenbreddels (changes requested):
small comments, let me know if they make sense

solara/website/pages/api/use_memo.md
iisakkirotko: Not to be confused indeed.

maartenbreddels:
Thank you 🙏

PR #341, opened 2023-10-25
Inputtextarea

draperjames:
This PR introduces the InputTextArea Widget

PR #376, opened 2023-11-13
feat: allow any `update_events` on `InputText`

iisakkirotko:
Previously `on_value` would be triggered whenever enter was pressed or `InputText` lost focus. With this PR we would allow people to modify these events.

### DIFF
--- a/solara/components/input.py
+++ b/solara/components/input.py
@@ -11,7 +11,7 @@ from solara.alias import rv as v
 T = TypeVar("T")
 
 
-def use_change(el: reacton.core.Element, on_value: Callable[[Any], Any], enabled=True):
+def use_change(el: reacton.core.Element, on_value: Callable[[Any], Any], enabled=True, enter_only=False):
     """Trigger a callback when a blur events occurs or the enter key is pressed."""
     on_value_ref = solara.use_ref(on_value)
     on_value_ref.current = on_value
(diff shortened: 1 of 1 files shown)


iisakkirotko:
* master
  * **PR #376** <a href="https://app.graphite.dev/github/pr/widgetti/solara/376?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/376?utm_source=stack-comment).

iisakkirotko:
> I think we should have this feature, but I wonder if we should have a different API. What about update_events=["blur", "keyup.enter"] as default arguments? This also allows a user to use "keyup.shift.enter" for instance. @mariobuikhuizen ?

Sounds like a good idea. We could then also support an arbitrary number of events as triggers. We should then also properly document the bug that causes only one event of a type to be picked up (unless that was fixed and I've forgotten).

maartenbreddels (changes requested):
I think we don't need the overload anymore, and the docstring needs change.

Weird btw, that mypy didn't complain about the overload arguments not matching the function arguments.

iisakkirotko:
> I think we don't need the overload anymore, and the docstring needs change.

Should be fixed now

maartenbreddels:
I think this is very useful for https://github.com/widgetti/solara/pull/341 as well, such that we can bind to shift-enter

PR #382, opened 2023-11-15
fix: do not use excessive memory when rendering markdown

maartenbreddels:
Due to legacy reasons we enabled the emoji plugin from pymdown.
Every time the Markdown object gets created, it will load the
emoji extension and will deep copy the emoji database.
This is about 150kb of data:
len(str(pymdownx.emoji1_db.emoji))

This is not a lot, but it is unnecessary and can be avoided.

### DIFF
--- a/solara/components/markdown.py
+++ b/solara/components/markdown.py
@@ -7,6 +7,7 @@ import warnings
 from typing import Any, Dict, List, Union
 
 import ipyvuetify as v
+import pymdownx.emoji
 import pymdownx.highlight
 import pymdownx.superfences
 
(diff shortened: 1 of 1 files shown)


maartenbreddels:
* master
  * **PR #382** <a href="https://app.graphite.dev/github/pr/widgetti/solara/382?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/382?utm_source=stack-comment).

PR #384, opened 2023-11-16
feat: Chat components, and AI example of use

iisakkirotko:


iisakkirotko:
* master
  * **PR #384** <a href="https://app.graphite.dev/github/pr/widgetti/solara/384?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/384?utm_source=stack-comment).

maartenbreddels (changes requested):
Exciting, starting to look really good!

The comments on solara/website/pages/examples/ai/chatbot.py I'm not so sure about, I feel a bit torn there. I think on the one hand it shows the flexibility of these components and solara, while at the same time the code may be a bit overwhelming... i think we need to digest this a bit and maybe do another call on this.

maartenbreddels (changes requested):
If you can add style and classes arguments, and we can merge :)

solara/lab/components/chat.py
```diff
@@ -0,0 +1,198 @@
...
+from typing_extensions import Literal
+
+import solara
+from solara.components.input import use_change
+
+
+@solara.component
+def ChatBox(
+    children: List[solara.Element] = [],
+    style: Optional[Union[str, Dict[str, str]]] = None,
```
maartenbreddels, line 13:
```suggestion
    style: Optional[Union[str, Dict[str, str]]] = {"flex-grow": "1", "flex-direction": "column-reverse", "overflow-y": "auto"},
```

solara/lab/components/chat.py
```diff
@@ -0,0 +1,198 @@
...
+    * `style`: CSS styles to apply to the component. Either a string or a dictionary.
+    * `classes`: A list of CSS classes to apply to the component.
+    """
+    style_flat = solara.util._flatten_style(style)
+
+    classes += ["chat-box"]
+    with solara.Column(
+        style={"flex-grow": "1", "flex-direction": "column-reverse", "overflow-y": "auto"},
+        classes=classes,
+        style_=style_flat,
```
maartenbreddels, line 33:
```suggestion
        classes=classes,
        style_=style_flat,
```

I thought this would make it more visible what the style is that we use.

solara/lab/components/chat.py
```diff
@@ -0,0 +1,198 @@
...
+    ):
+        for child in list(reversed(children)):
+            solara.display(child)
+
+
+@solara.component
+def ChatInput(
+    send_callback: Optional[Callable] = None,
+    disabled: bool = False,
+    style: Optional[Union[str, Dict[str, str]]] = None,
```
maartenbreddels, line 43:
```suggestion
    style: Optional[Union[str, Dict[str, str]]] = {"align-items": "center"},
```

solara/lab/components/chat.py
```diff
@@ -0,0 +1,198 @@
...
+    * `disabled`: Whether the input should be disabled. Useful for disabling sending further messages while a chatbot is replying,
+        among other things.
+    * `style`: CSS styles to apply to the component. Either a string or a dictionary. These styles are applied to the container component.
+    * `classes`: A list of CSS classes to apply to the component. Also applied to the container.
+    """
+    message, set_message = solara.use_state("")  # type: ignore
+    style_flat = solara.util._flatten_style(style)
+
+    if "align-items" not in style_flat:
+        style_flat += "align-items: center;"
```
maartenbreddels, line 61:
```suggestion
    style_flat = solara.util._flatten_style(style)
```

same here

solara/lab/components/chat.py
```diff
@@ -0,0 +1,198 @@
...
+        style_flat += f"border-radius: {border_radius if border_radius is not None else ''};"
+    if f"border-top-{'right' if user else 'left'}-radius" not in style_flat:
+        style_flat += f"border-top-{'right' if user else 'left'}-radius: 0;"
+    if "padding" not in style_flat:
+        style_flat += "padding: .5em 1.5em;"
+
+    msg_uuid = solara.use_memo(lambda: str(uuid.uuid4()), dependencies=[])
+    with solara.Row(
+        justify="end" if user else "start",
+        style={"flex-direction": "row-reverse" if user else "row", "padding": "5px"},
```
maartenbreddels, line 132: Does it make sense to not apply the style at the top level container like we always do?
It's a bit tricky how we deal with the default style that is condition..

solara/lab/components/chat.py
iisakkirotko: I would do this one the same way as the others (conditionally). I think users might find it confusing if they aren't familiar with CSS and want to change something like `{"background-color": "green"}`, and that makes the container not work. Let me know what you think.

maartenbreddels: True, lets keep it this way.

solara/lab/components/chat.py
maartenbreddels: Ok, we decided to do it like this for now, it's in lab, so it can always change if we feel limited by it.

PR #435, opened 2023-12-30
Add support for customized authentication for fastAPI and starlett

warm200:
Propose a customized way to do authentication for fastAPI and starlette

the reason is nowadays some companies may already have their own solution other than existing Oauth providers currently supported in solara.
[Auth0](https://auth0.com/) and [Fief](https://fief.dev/)
this is to make it more flexible and give user options to implement their own authentication method.

## Usage:

`export AUTH_MODULE_PATH=file/to/your/authenticate_module.py`

the module need to have a method called `authenticate` and return is bool.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ dev = [
     "types-markdown",
     "types-PyYAML",
     "pytest",
+    "pytest-asyncio",
     "pytest-mock",
     "pytest-cov",
     "pytest-timeout",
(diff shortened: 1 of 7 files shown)


maartenbreddels:
Hi Wallace,

Thank you for contributing to Solara!
I am not sure this is needed in solara, since it's a starlette specific PR, and there is already a starlette specific way this can be done I think:
A custom middleware can be added following:
https://www.starlette.io/authentication/
To the app object, following:
https://solara.dev/docs/deploying/self-hosted

Let me know if this can solve your problem, and maybe this PR can be replaced by adding this to the documentation?

Regards,

Maarten

maartenbreddels:
I'll close this for now, feel free to re-open if you feel this PR really adds something compared to what we already have.

PR #477, opened 2024-01-22
fix: avoid extra scrollbar when having no appbar or sidebar

maartenbreddels:
This issue was introduced in 3884d6a035ab11098d78f6ff6c76aec060dfd785 and was caused by a full height 100% chain and a single element with padding, causing the solara-autorouter-content to be pushed down and creating scrollbars on its parents.
If instead, we add the scrollbar to the solara-autorouter-content element, we can avoid this issue.

See
image

where the anonymous div is causing scrollbars, and the solara-autorouter-content causes overflow.

The v-input has a top-margin of 4, triggering this.

After this fix, we don't see overflow and scrollbars:
image

### DIFF
--- a/solara/server/assets/style.css
+++ b/solara/server/assets/style.css
@@ -112,6 +112,7 @@ div.highlight {
 
 .solara-autorouter-content {
     height: 100%;
+    overflow: auto;
 }
 
 /* originally from index.css */


solara/server/assets/style.css
```diff
@@ -112,6 +112,7 @@ div.highlight {
 
 .solara-autorouter-content {
     height: 100%;
+    overflow: auto;
```
iisakkirotko, line 115: This seems to cause a different issue:
image
Another, more different scroll bar

iisakkirotko: The cause seems to be the fact that the 8px margins on the `Card`, together with height/width 100% push it beyond the bounds of the element containing it...

iisakkirotko: Also causes this to happen with `.markdown-output` cells
image

iisakkirotko:
A soft way to solve this would be to add some conditionality to the `-12px` margins,

```CSS
.row.solara-content-main{
    margin: 0;
}
.row.solara-autorouter-content{
    margin: 0;
}
```

Another approach would be to override the margins alltogether

```CSS
div.row{ /* add div for specificity */
    margin: 0;
}
````

Issue #486, opened 2024-01-29
Altair and other JS extensions do not work in Voila or Colab


maartenbreddels:
In https://github.com/widgetti/solara/blob/master/solara/widgets/vue/vegalite.vue
we assume the `/_solara/_cdn/` endpoint can be reached. This endpoint exists in notebook, lab and solara server, but not in Voila (except when running as a voila extension in jupyter server) or colab.

A workaround for now is to manually set the `solara_cdn` javascript variable.
```python
import IPython.display
# make solara use a real cdn, since the solara cdn proxy is not available
display(IPython.display.Javascript("solara_cdn = 'https://cdn.jsdelivr.net/npm/'"))
```

maartenbreddels:
Found this our while writing a test for https://github.com/widgetti/solara/pull/479

iisakkirotko:
Also worth noting that the endpoint is available in VSCode

maartenbreddels:
Altair should work now due to #487, other js libraries not yet

Jhsmit:
For me Altair also doesnt work in notebooks

`404 GET /_solara/cdn/requirejs@2.3.6/require.js`

maartenbreddels:
Should be fixed due to #479 

Keeping this open till we fix this for the other libraries.

maartenbreddels:
Fixed in #763  (Solara 1.39.0)

PR #480, opened 2024-01-24
fix: Mermaid and Math rendering in Jupyter Lab and VSCode

iisakkirotko:
Should fix https://github.com/widgetti/solara/issues/175, and partially fix https://github.com/widgetti/solara/issues/446

PR #479, opened 2024-01-24
fix: solara.server.jupyter.server_extension not activating

iisakkirotko:
I can't say I figured out what broke this

PR #487, opened 2024-01-29
refactor: split assets settings in server and non-server parts

maartenbreddels:
Based on:
 * https://github.com/widgetti/solara/pull/479
 * https://github.com/widgetti/solara/pull/480
 * https://github.com/widgetti/solara/issues/486

We decided that using the proxy and the cdn if the proxy is not used
is not only something to be used in solara server. In vscode
or colab, we cannot use the proxy, and need to use the cdn directly.

### DIFF
--- a/solara/server/cdn_helper.py
+++ b/solara/server/cdn_helper.py
@@ -4,7 +4,7 @@ import shutil
 
 import requests
 
-from . import settings
+import solara.settings
 
 logger = logging.getLogger("Solara.cdn")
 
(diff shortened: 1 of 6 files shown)


maartenbreddels:
### Merge activity

* **Jan 30, 4:33 AM**: @maartenbreddels started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/487).

PR #515, opened 2024-02-21
feat: decrease chances of reactive key collisions

iisakkirotko:
A more permanent fix (although still not foolproof) for https://github.com/widgetti/solara/issues/510. Also take a look at https://github.com/widgetti/solara/pull/514

PR #517, opened 2024-02-22
feat: improve logging of reactive variable changes

iisakkirotko:
The way that the `log_level`-settings is set is kind of hacky right now, but we need to discuss implementing a real solution in a different way

### DIFF
--- a/solara/__main__.py
+++ b/solara/__main__.py
@@ -367,7 +367,10 @@ def run(
 
     if log_level is not None:
         LOGGING_CONFIG["loggers"]["solara"]["level"] = log_level.upper()
+        settings.main.log_level = log_level.upper()
         # LOGGING_CONFIG["loggers"]["reacton"]["level"] = log_level.upper()
+    else:
+        settings.main.log_level = "ERROR"
 
     log_level = log_level_uvicorn
     del log_level_uvicorn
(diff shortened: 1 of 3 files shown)


iisakkirotko:
So testing the performance, the cost is only at start-up of the server, but with this change it takes about 4s longer to start the server than without. I'll take a look at what part of the change requires so much time next week.

maartenbreddels:
Btw, i except this PR to build on top of #515 ? Maybe a squash issue?

TODO: 
 * [ ] find out why it takes so much more time (Iisakki)
 * [ ]  find out why it seems to initialize twice (Maarten)

iisakkirotko:
> Btw, I expected this PR to build on top of #515 ? Maybe a squash issue?

To me these changes feel unrelated (even though they both use `inspect.stack()`), since one is about logging, while the other is fixing an issue. You could also argue that #515 could be based on this one, and make use of `var._varname`.

iisakkirotko:
The performance of various ways to get caller names was tested [here](https://gist.github.com/JettJones/c236494013f22723c1822126df944b12), from which it's pretty fair to say that we chose the slowest way to do this initially. I'll take a look at using a rewriting this PR using one of the better performing methods.

PR #523, opened 2024-02-26
fix: invalidate cache on code update

iisakkirotko:


iisakkirotko:
* **#523** <a href="https://app.graphite.dev/github/pr/widgetti/solara/523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/523?utm_source=stack-comment).

maartenbreddels (changes requested):
Code looks good, but as discussed, needs to hash all the files in the directory, since index.js (for instance) is more likely to change compared to extension.js

iisakkirotko:
Did a little shuffle with the base branch to get rid of github showing extra commits in the PR (https://github.com/widgetti/solara/commit/5f3946fa783810b67406853d49cf24baced7b122 and https://github.com/widgetti/solara/commit/7bf2ce53c7ede6eec3883abdc7d5f2625e22ccd1 were shown in the PR)

solara/util.py
```diff
@@ -257,9 +257,15 @@ def parse_timedelta(size: str) -> float:
...
+def get_file_hash(path: Union[Path, Generator], algorithm="md5") -> Tuple[bytes, str]:
     """Compute the hash of a file. Note that we also return the file content as bytes."""
-    data = path.read_bytes()
+    if isinstance(path, Generator):
+        data = b""
+        for file in path:
+            chunk = file.read_bytes()
+            data += chunk
+    else:
+        data = path.read_bytes()
```
maartenbreddels, line 268: I had the same idea of accepting a Generator (can it be property typed? Generate[Path] orso?).
But it's more efficient to first create the hash object, and call .update on it, see
https://docs.python.org/3/library/hashlib.html#usage

solara/server/templates/solara.html.j2
```diff
@@ -397,7 +398,16 @@
         'jupyter-vuetify': 'nbextensions/jupyter-vuetify/nodeps',
             {% endif -%}
                     },
-                }
+                },
+                urlArgs: function (id, url) {
+                    id = id.replace("/static/nbextensions/", "").replace(".js", "");
+                    if (nbextensionHashes[id] !== undefined) {
+                        console.log('nbextensionHashes', id, nbextensionHashes[id]);
+                        return (url.indexOf('?') === -1 ? '?' : '&') + nbextensionHashes[id];
```
maartenbreddels, line 406: maybe add a comment why we do this

maartenbreddels (changes requested):

solara/util.py
iisakkirotko: Yeah, from a MyPy error I got it seemed to be `Generator[Path, None, None]`.

maartenbreddels (approved):
Looks perfect 👍

PR #557, opened 2024-03-14
fix: load theme earlier when using SSG

iisakkirotko:


solara/server/templates/solara.html.j2
```diff
@@ -38,9 +36,22 @@
...
+        border-color: #fff;
+        color: rgba(0,0,0,.87)
+    }
+
+    .theme--dark .v-sheet {
+        background-color: #1e1e1e;
+        border-color: #1e1e1e;
+        color: #fff
+    }
+</style>
```
iisakkirotko, line 54: @maartenbreddels what do you think, should these options be configurable?

solara/server/templates/solara.html.j2
```diff
@@ -175,6 +186,43 @@
...
+        let app_container = document.getElementById('app');
+        if (inDarkMode()) {
+            app_container.classList.remove('theme--light');
+            app_container.classList.add('theme--dark');
+        } else {
+            app_container.classList.remove('theme--dark');
+            app_container.classList.add('theme--light');
+        }
+        async function changeThemeCSS(theme) {
+            let css = await fetch(`/static/assets/theme-${theme}.css`).then(r => r.text());
```
maartenbreddels, line 221: This fails on pycafe where rootPath is `/_app`
```suggestion
            let css = await fetch(`${solara.rootPath}/static/assets/theme-${theme}.css`).then(r => r.text());
```

solara/server/templates/solara.html.j2
```diff
@@ -27,7 +25,7 @@
     {{ resources.include_css("/static/highlight.css") }}
     {{ resources.include_css("/static/highlight-dark.css") }}
     {{ resources.include_css("/static/assets/style.css") }}
-    <style id="jupyter-theme-css">
+    <style id="jupyter-theme-css" class="solara-template-css">
         {{ resources.include_css("/static/assets/theme-"~theme.variant~".css") }}
```
maartenbreddels, line 29: Do we want this content there? Since we immediately fill it using `changeThemeCSS`, and it might result in an extra flash?

solara/server/templates/solara.html.j2
```diff
@@ -175,6 +186,43 @@
...
+        function inDarkMode() {
+            if (getThemeVariant() == 'auto') {
+                return prefersDarkScheme();
+            }
+            else {
+                return getThemeVariant() == 'dark';
+            }
+        }
+        // Init theme
+        let app_container = document.getElementById('app');
```
maartenbreddels, line 212: camelCasing?
```suggestion
        let appContainer = document.getElementById('app');
```

solara/server/templates/solara.html.j2
maartenbreddels: Can you explain why it is there in a comment, and why we remove it later on?
If I think I understand (first render) than this is fine for now I think.

maartenbreddels (changes requested):
Looking good!

solara/server/templates/solara.html.j2
iisakkirotko: Good idea! I think that way we save one load of the stylesheet.

packages/solara-enterprise/solara_enterprise/ssg.py
```diff
@@ -205,6 +205,8 @@ def _ssg_data(html: str) -> Optional[SSGData]:
     rendered_styles = soup.find_all("style")
     for style in rendered_styles:
         style_html = str(style)
+        if 'class="solara-template-css"' in style_html:
+            continue
```
maartenbreddels, line 209:
```suggestion
        # skip css that was already in the template so we don't include it multiple times
        # or such that we do not include the CSS from the theme as ssg build time
        if 'class="solara-template-css"' in style_html:
            continue
```

maartenbreddels (approved):
Looks good 👍  just a final clarification for future us.

Issue #260, opened 2023-08-30
Multiple file support for FileDrop


hkayabilisim:
Currently, FileDrop only supports single file upload. If you drag multiple files, it only processes one of them. Is there a plan to implement multiple file support? If you can provide some tips, I can also try to implement it.

maartenbreddels:
Hi,

this is not on our list yet, but would be nice to have, do you agree @mariobuikhuizen ? I think @mariobuikhuizen also once added support for directories right?
There is also https://github.com/widgetti/solara/pull/46
Both are built on top of https://github.com/widgetti/ipyvuetify/tree/master/ipyvuetify/extra which should support multiple files.
a PR would we great!

Regards,

Maarten

hkayabilisim:
I checked the file_drop.py and file_drop.vue. And I got the impression that multiple file support is almost there. I tried to modify them but couldn't make it properly. I'm in a hurry, so, I'm moving on to using multiple FileDrops. If I find time, I will get back and give it a second try.

hkayabilisim:
I've implemented this feature: https://github.com/widgetti/solara/pull/562

PR #421, opened 2023-12-11
refactor: Uniting API, Docs, and Examples under Documentation on solara.dev

iisakkirotko:
refactor: United API, Examples, and Docs under Documentation page

PR #562, opened 2024-03-16
feat: multiple file support for file_drop

hkayabilisim:
Added multiple file support to file_drop component.  
Same as the original component, directories are simply ignored. Multiple file support was actually embedded in the underlying VUE scripts. I've done minor changes. 

**Compatibility**:
on_file callback function now accepts List[FileInfo] as opposed to single FileInfo object. This may bring some problems for existing codes. As a remedy, if the number of files is just one, we can return a single FileInfo. If not, we return list. I leave it up to you to decide.

This is related to: https://github.com/widgetti/solara/issues/260

### DIFF
--- a/solara/components/file_drop.py
+++ b/solara/components/file_drop.py
@@ -1,6 +1,6 @@
 import threading
 import typing
-from typing import Callable, Optional, cast
+from typing import Callable, List, Optional, Union, cast
 
 import traitlets
 from ipyvue import Template
@@ -25,45 +25,26 @@ class FileDropZone(FileInput):
     template_file = (__file__, "file_drop.vue")
     items = traitlets.List(default_value=[]).tag(sync=True)
     label = traitlets.Unicode().tag(sync=True)
+    multiple = traitlets.Bool(True).tag(sync=True)
 
 
 @solara.component
-def FileDrop(
-    label="Drop file here",
+def _FileDrop(
+    label="Drop file(s) here",
     on_total_progress: Optional[Callable[[float], None]] = None,
-    on_file: Optional[Callable[[FileInfo], None]] = None,
+    on_file: Optional[Callable[[Union[FileInfo, List[FileInfo]]], None]] = None,
     lazy: bool = True,
+    multiple: bool = False,
 ):
-    """Region a user can drop a file into for file uploading.
+    """Generic implementation used by FileDrop and FileDropMultiple.
 
-    If lazy=True, no file content will be loaded into memory,
-    nor will any data be transferred by default.
-    A file object is passed to the `on_file` callback, and data will be transferred
-    when needed.
-
-    If lazy=False, the file content will be loaded into memory and passed to the `on_file` callback via the `.data` attribute.
-
-    The on_file callback takes the following argument type:
-    ```python
-    class FileInfo(typing.TypedDict):
-        name: str  # file name
-        size: int  # file size in bytes
-        file_obj: typing.BinaryIO
-        data: Optional[bytes]: bytes  # only present if lazy=False
-    ```
-
-
-    ## Arguments
-     * `on_total_progress`: Will be called with the progress in % of the file upload.
-     * `on_file`: Will be called with a `FileInfo` object, which contains the file `.name`, `.length` and a `.file_obj` object.
-     * `lazy`: Whether to load the file content into memory or not. If `False`,
-        the file content will be loaded into memory and passed to the `on_file` callback via the `.data` attribute.
-
+    If multiple=True, multiple files can be uploaded.
     """
+
     file_info, set_file_info = solara.use_state(None)
     wired_files, set_wired_files = solara.use_state(cast(Optional[typing.List[FileInfo]], None))
 
-    file_drop = FileDropZone.element(label=label, on_total_progress=on_total_progress, on_file_info=set_file_info)  # type: ignore
+    file_drop = FileDropZone.element(label=label, on_total_progress=on_total_progress, on_file_info=set_file_info, multiple=multiple)  # type: ignore
 
     def wire_files():
         if not file_info:
@@ -83,14 +64,76 @@ def FileDrop(
         if not wired_files:
             return
         if on_file:
-            if not lazy:
-                wired_files[0]["data"] = wired_files[0]["file_obj"].read()
+            for i in range(len(wired_files)):
+                if not lazy:
+                    wired_files[i]["data"] = wired_files[i]["file_obj"].read()
+                else:
+                    wired_files[i]["data"] = None
+            if multiple:
+                on_file(wired_files)
             else:
-                wired_files[0]["data"] = None
-            on_file(wired_files[0])
+                on_file(wired_files[0])
 
     result: solara.Result = hooks.use_thread(handle_file, [wired_files])
     if result.error:
         raise result.error
 
     return file_drop
+
+
+@solara.component
+def FileDrop(
+    label="Drop file here",
+    on_total_progress: Optional[Callable[[float], None]] = None,
+    on_file: Optional[Callable[[FileInfo], None]] = None,
+    lazy: bool = True,
+):
+    """Region a user can drop a file into for file uploading.
+
+    If lazy=True, no file content will be loaded into memory,
+    nor will any data be transferred by default.
+    If lazy=False, file content will be loaded into memory and passed to the `on_file` callback via the `FileInfo.data` attribute.
+
+
+    A file object is of the following argument type:
+    ```python
+    class FileInfo(typing.TypedDict):
+        name: str  # file name
+        size: int  # file size in bytes
+        file_obj: typing.BinaryIO
+        data: Optional[bytes]: bytes  # only present if lazy=False
+    ```
+
+
+    ## Arguments
+     * `on_total_progress`: Will be called with the progress in % of the file upload.
+     * `on_file`: Will be called with a `FileInfo` object, which contains the file `.name`, `.length` and a `.file_obj` object.
+     * `lazy`: Whether to load the file contents into memory or not. If `False`,
+        the file contents will be loaded into memory via the `.data` attribute of file object(s).
+
+    """
+
+    return _FileDrop(label=label, on_total_progress=on_total_progress, on_file=on_file, lazy=lazy, multiple=False)
+
+
+@solara.component
+def FileDropMultiple(
+    label="Drop files here",
+    on_total_progress: Optional[Callable[[float], None]] = None,
+    on_file: Optional[Callable[[List[FileInfo]], None]] = None,
+    lazy: bool = True,
+):
+    """Region a user can drop multiple files into for file uploading.
+
+    Almost identical to `FileDrop` except that multiple files can be dropped and `on_file` is called
+    with a list of `FileInfo` objects.
+
+    ## Arguments
+     * `on_total_progress`: Will be called with the progress in % of the file(s) upload.
+     * `on_file`: Will be called with a `List[FileInfo]`.
+        Each `FileInfo` contains the file `.name`, `.length`, `.file_obj` object, and `.data` attributes.
+     * `lazy`: Whether to load the file contents into memory or not.
+
+    """
+
+    return _FileDrop(label=label, on_total_progress=on_total_progress, on_file=on_file, lazy=lazy, multiple=True)
--- a/solara/components/file_drop.vue
+++ b/solara/components/file_drop.vue
@@ -1,6 +1,18 @@
 <template>
   <div ref="dropzone" class="solara-file-drop" effectAllowed="move">
-    {{ (file_info && file_info.length === 1) ? file_info[0].name : label }}
+    <template v-if="file_info && file_info.length > 0">
+      <template v-if="multiple">
+        <div v-for="file in file_info">
+          {{ file.name }}
+        </div>
+      </template>
+      <template v-else>
+        {{ file_info[0].name }}
+      </template>
+    </template>
+    <template v-else>
+      {{ label }}
+    </template>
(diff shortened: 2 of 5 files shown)


solara/components/file_drop.vue
```diff
@@ -1,6 +1,13 @@
 <template>
   <div ref="dropzone" class="solara-file-drop" effectAllowed="move">
-    {{ (file_info && file_info.length === 1) ? file_info[0].name : label }}
+    <div v-if="file_info && file_info.length">
```
mariobuikhuizen, line 3: Ideally we would also keep the HTML the same for the single file case, as users might have CSS rules that depend on that structure. You can use `<template v-if=`, this wil not produce an HTML element in the output.

solara/components/file_drop.vue
```diff
@@ -1,6 +1,13 @@
...
+      <ul>
+        <li v-for="file in file_info">{{ file.name }}  {{ file.size }} bytes</li>
```
mariobuikhuizen, line 5: I would like the list to not have bullet points.

mariobuikhuizen (changes requested):
Thanks for your contribution @hkayabilisim!

We would like to keep this component backward compatible. We could accomplish this by adding an argument `multiple` with a default value of `False` and have on_file also work with a single file (see [this](https://github.com/widgetti/solara/blob/e3ea88aa617b30da89d8fe92682c7346065d70e4/solara/components/input.py#L141) for an example of the typing, note the argument `optional: Literal[False]`)

solara/components/file_drop.vue
```diff
@@ -16,16 +23,17 @@ module.exports = {
...
+      this.native_file_info = nativeFiles
       this.file_info = this.native_file_info.map(
-          ({name, isFile, size}) => ({
-            name,
-            isFile,
-            size,
-          }));
+        ({name, isFile, size}) => ({
+          name,
+          isFile: !!isFile, // Ensure this is a boolean
```
mariobuikhuizen, line 34: Why is this needed? The spec says it's a boolean: https://developer.mozilla.org/en-US/docs/Web/API/FileSystemEntry/isFile

solara/components/file_drop.vue
hkayabilisim: I guess, this one will solve the problem.

~~~vue
<template>
  <div ref="dropzone" class="solara-file-drop" effectAllowed="move">
    <template v-if="file_info && file_info.length === 1"> 
      {{ file_info[0].name }}
    </template>
    <template v-else-if="file_info && file_info.length > 1"> 
      <div v-for="file in file_info">
        {{ file.name }}
      </div>
    </template>
    <template v-else> 
      {{ label }}
    </template>
  </div>
</template>
~~~

solara/components/file_drop.vue
hkayabilisim: Done!

solara/components/file_drop.vue
hkayabilisim: Yes `FileSystemEntry` has `isFile` property. However,  after applying Promises (which effectively creates File objects), we don't have isFile property anymore. So isFile of the returned object was undefined. This was true for the original version. By using !!isFile, I was trying to create a boolean variable from undefined. Anyway, it turns out that isFile is not important because it is never used after this place. So this bug was not causing any problem. 

I'm suggesting the following. I think it is ok to set isFile=true because in the beginning we already filter files.

~~~javascript
this.file_info = this.native_file_info.map(
  ({name, size}) => ({
    name,
    isFile: true,
    size,
  }));
~~~

hkayabilisim (commented):
missing ';' at the end of the "padding:" line was causing CSS problem.

hkayabilisim (commented):
Changes required for backward compatibility:

1. multiple property is added to FileDrop
2. Two different on_file callback signature used: FileInfo and List[FileInfo]
3. FileDrop is overloaded to match two different on_file signatures.
4. FileDrop apidoc is updated
5. FileDrop example is updated
6. bullet list in VUE template is removed
7. isFile property is always set to True (this was a hidden bug)
8. app/scatter.py is reverted to original version

hkayabilisim (commented):
Comments of @mariobuikhuizen are handled in [7d44858ec65c103c299d9757c4ef169e5f5b9f9b](https://github.com/widgetti/solara/pull/562/commits/7d44858ec65c103c299d9757c4ef169e5f5b9f9b)

maartenbreddels:
Looking good!
A few issues

- due to a large refactor, files have moved (see https://github.com/widgetti/solara/pull/421)
- I think we need to pass 'multiple' to the frontend, because now you can still drag in multiple files and they render in the component, I think we should show only 1 
- Some style issues (if you install pre-commit and run `pre-commit install`, those will be fixed for you - and we should document this on https://solara.dev/docs/howto/contribute )

Looking forward to see this green in CI, and get this merged 🥳

hkayabilisim:
Used "multiple" option in file_drop.vue to show single file when multiple is not selected.

hkayabilisim:
Hi @maartenbreddels 

There is only one error I couldn't resolve. Help needed!

image

iisakkirotko:
Hi @hkayabilisim! Again, thanks for the contribution. 

The remaining issue is due to the how the component is used in the documentation: `multiple_upload` is a variable, and therefore not a Literal, so it doesn't match any of the overload variants. This can be fixed by adding an overload variant which takes a `bool` instead of `Literal[...]` for `multiple`, i.e. 

```python
@overload
@solara.component
def FileDrop(
    label=...,
    on_total_progress: Optional[Callable[[float], None]] = ...,
    on_file: Optional[Callable[[Union[FileInfo, List[FileInfo]]], None]] = ...,
    lazy: bool = ...,
    multiple: bool = ...,
) -> ipyvuetify.extra.FileInput:
    ...
```

Then the two functions that are currently used in the documentation could be combined to one that takes a Union of `FileInfo` and `List[FileInfo]` as an argument, like so:

```python
def process_files(files: Union[FileInfo, List[FileInfo]]):
    if not isinstance(files, list):
        files = [files]
    set_filename([f["name"] for f in files])
    set_size([f["size"] for f in files])
    set_content([f["file_obj"].read(100) for f in files])
```

solara/components/file_drop.py
```diff
@@ -25,25 +26,51 @@ class FileDropZone(FileInput):
...
+@overload
+@solara.component
+def FileDrop(
+    label: str = ...,
+    on_total_progress: Optional[Callable[[float], None]] = ...,
+    on_file: Optional[Callable[[List[FileInfo]], None]] = None,
+    lazy: bool = ...,
+    multiple: Literal[True] = ...,
+) -> ipyvuetify.extra.FileInput:
+    ...
```
iisakkirotko, line 54: Hi @hkayabilisim! Again, thanks for the contribution. 

The remaining issue is due to the how the component is used in the documentation: `multiple_upload` is a variable, and therefore not a Literal, so it doesn't match any of the overload variants. This can be fixed by adding an overload variant which takes a `bool` instead of `Literal[...]` for `multiple`, i.e. 

```suggestion
@overload
@solara.component
def FileDrop(
    label=...,
    on_total_progress: Optional[Callable[[float], None]] = ...,
    on_file: Optional[Callable[[Union[FileInfo, List[FileInfo]]], None]] = ...,
    lazy: bool = ...,
    multiple: bool = ...,
) -> ipyvuetify.extra.FileInput:
    ...
```

solara/website/pages/documentation/components/input/file_drop.py
```diff
@@ -10,25 +11,32 @@
...
-        set_size(file["size"])
-        f = file["file_obj"]
-        set_content(f.read(100))
+    def process_multiple_files(files: List[FileInfo]):
+        set_filename([f["name"] for f in files])
+        set_size([f["size"] for f in files])
+        set_content([f["file_obj"].read(100) for f in files])
+
+    def process_single_file(file: FileInfo):
+        process_multiple_files([file])
```
iisakkirotko, line 25: Then the two functions that are currently used in the documentation could be combined to one that takes a Union of `FileInfo` and `List[FileInfo]` as an argument, like so:

```suggestion
def process_files(files: Union[FileInfo, List[FileInfo]]):
    if not isinstance(files, list):
        files = [files]
    set_filename([f["name"] for f in files])
    set_size([f["size"] for f in files])
    set_content([f["file_obj"].read(100) for f in files])
```

solara/website/pages/documentation/components/input/file_drop.py
```diff
@@ -10,25 +11,32 @@
...
 
     with solara.Div() as main:
+        solara.Checkbox(label="Multiple upload", value=multiple_upload, on_value=set_multiple_upload)
         solara.FileDrop(
-            label="Drag and drop a file here to read the first 100 bytes",
-            on_file=on_file,
+            label="Drag and drop files(s) here to read the first 100 bytes.",
+            on_file=process_multiple_files if multiple_upload else process_single_file,
```
iisakkirotko, line 31: And

```suggestion
            on_file=process_files,
```

maartenbreddels:
Hi Huseyin,

we had quite a long discussion internally on this PR.
For ToggleButton https://github.com/widgetti/solara/blob/master/solara/components/togglebuttons.py we had two different components
 * ToggleButtonsSingle
 * ToggleButtonsMultiple

This is because otherwise the typing becomes quite complex (as you can see now for FileDrop), and it only works for literals. So if you want to take advantage of the typing you need to do:
```
if multiple:
  FileDrop(..., multiple=True)
else:
  FileDrop(..., multiple=False)
```

Which is a bit odd. For this reason, it might be better to have
`FileDrop` and `FileDropMultiple` (they can call the same internal component btw, it's just the public API)

However, also that would be inconsistent in the naming, so either solution is not perfect.

Two questions:
 1. What do you think about the typing and one vs two components
 2. If you also prefer FileDropMultiple, are you willing to do the changes? If not, we're happy to help.

Let us know, and thanks you for your help on this.

Regards,

Maarten

hkayabilisim:
Hi Maarten,

Getting rid of some complexity by creating two variants (FileDrop and FileDropMultiple) seems a good idea. 

I've just created FileDropMultiple and commited to the branch and tested little bit (https://github.com/widgetti/solara/pull/562/commits/fa6cd2367b445d674e4d89ccc827f386d60bf7ac). It seems ok, and CI checks are green. Yet, you should revise my implementation in case if I missed some points. 

Sincerely
Huseyin

maartenbreddels:
Hi Huseyin,

whow, many thank for doing all this work!
CI looks green (ignore that failing one, that's not related to what you do)
Will review after the weekend probably. Have a good weekend.

Regards,

Maarten

solara/components/file_drop.py
```diff
@@ -33,17 +35,15 @@ def FileDrop(
     on_total_progress: Optional[Callable[[float], None]] = None,
     on_file: Optional[Callable[[FileInfo], None]] = None,
     lazy: bool = True,
-):
+) -> ipyvuetify.extra.FileInput:
```
maartenbreddels, line 38:
```suggestion
)
```
This is not the case, the function does not return a widget, but the type inference will (or should) figure this out, so no need to add

solara/components/file_drop.py
```diff
@@ -94,3 +94,59 @@ def handle_file(cancel: threading.Event):
...
+def FileDropMultiple(
+    label="Drop files here",
+    on_total_progress: Optional[Callable[[float], None]] = None,
+    on_file: Optional[Callable[[List[FileInfo]], None]] = None,
+    lazy: bool = True,
+) -> ipyvuetify.extra.FileInput:
+    """Region a user can drop multiple files into for file uploading.
+
+    Almost identical to `FileDrop` except that `on_file` is called
+    with a list of `FileInfo` objects.
```
maartenbreddels, line 109:
```suggestion
    Almost identical to `FileDrop` except that multiple files can be dropped and `on_file` is called
    with a list of `FileInfo` objects.
```

solara/components/file_drop.py
```diff
@@ -94,3 +94,59 @@ def handle_file(cancel: threading.Event):
...
+                    wired_files[i]["data"] = wired_files[i]["file_obj"].read()
+                else:
+                    wired_files[i]["data"] = None
+            on_file(wired_files)
+
+    result: solara.Result = hooks.use_thread(handle_file, [wired_files])
+    if result.error:
+        raise result.error
+
+    return file_drop
```
maartenbreddels, line 152: I don't think we want this file duplication. I'm sorry if I wasn't clear in my previous message.
Basically what I suggested is that your previous implementation could be used by these two implementations. So lets say we call your previous component `_FileDrop`, we simply call `return _FileDrop(...., multiple=True)` from this component.

maartenbreddels (changes requested):
Awesome work.
Let me know if you have the energy to do these last changes. I know we're asking a lot from you, so we are happy to do the last mile.

hkayabilisim:
No problem Maarten! I've added _FileDrop to prevent code repetitions. I hope this is what you have in your mind!

hkayabilisim (commented):
It may not fit 100% to the code styling requirements but  I think my last commit https://github.com/widgetti/solara/pull/562/commits/e5e6f405fd6c48b4c33b2d43f7ad57ce757fefff solves most of the highlighted issues. 

One of the checks keeps failing but I couldn't understand why.

iisakkirotko:
Hi @hkayabilisim! Just wanted to let you know that the CI failure is still unrelated, it should be fixed by https://github.com/widgetti/solara/pull/574. We'll try to get this PR in soon™️ :)

maartenbreddels:
This branch had some flakiness that we wanted to attack (see the last commit) before merging it. Thanks for your patience!

PR #581, opened 2024-04-03
fix: close playwright properly when building ssg

iisakkirotko:
Previously we attempted to close the playwrights after the threadpool has been closed. However, outside of the async context we only find `pw.browser` (as well as all other properties) to be `None`, which previously resulted in an error due to us asserting that `pw.browser is not None`. Closing the browser and context manager properly should ensure they don't drain any resources after ssg generation is done.

I'm not sure if there is a concern that `close_playwrights` executes before the ssg is done, resulting in the opening of a new context, and that one not being closed.

### DIFF
--- a/packages/solara-enterprise/solara_enterprise/ssg.py
+++ b/packages/solara-enterprise/solara_enterprise/ssg.py
@@ -5,7 +5,7 @@ import time
 import typing
 import urllib
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 import solara
 from rich import print as rprint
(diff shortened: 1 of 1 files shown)


iisakkirotko:
* **#581** <a href="https://app.graphite.dev/github/pr/widgetti/solara/581?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

iisakkirotko:
We can also see that currently ssg fails due to the assertion error, see [the workflow run](https://github.com/widgetti/solara/actions/runs/8541187045)

Issue #599, opened 2024-04-11
feat: support ibis tables (...and define a more formal protocol)


NickCrews:
I would love to render (very very large, ie millions of rows) ibis tables in solara.DataTable. This currently doesn't work because

- solara calls `len(df)` on the data. ibis uses `.count()` instead, to make the execution very explicit
- the logic to go from ibis table to a list[dict] of records isn't implemented for ibis

I see a few ways to support this:

- manually add ibis support :)
- Make some more formal protocol how solara interacts with dataframes, and then let users implement the translation layer themselves.
- Use ibis itself as your dataframe abstraction layer :) Another dependency, doesn't support vaex, probably not worth it.

I'm cooking up a little PR right now that at least refactors things to make them a bit cleaner, regardless of if we do something more drastic. Thank you!

maartenbreddels:
Hi Nick,

we share a common dream then!

Our idea for the DataTable (or DataFrame) is to separate it into a very dumb view, that takes in the length and the records, etc. Our current DataTable would then use that dump view, so that others (like you) can easily build its own.
However, that is independent of #600 and I like that PR.
What do you think of this idea? 

Note there is also https://github.com/data-apis/dataframe-api

Also related to this is: https://codepen.io/eddie1952/pen/jObPvKO which would make it scrollable for very large lengths, although it would be limited to a fixed row height.

Regards,

Maarten

NickCrews:
If I understand correctly, you are suggesting making the DataTable component take arguments `records: list[dict], n: int` instead of the current `df: DataFrameLike`? I am a small -1 on this idea:
- if we show 0 rows, then we still need to know the schema of the table. So that would have to get passed too
- In the format callback, we pass in the original dataframe. People can do a lot more useful processing on a real DataFrame than a list[dict]
- If people have to do the to_records() themselves, they are going to make mistakes
- to do lazy row slicing, we would need to also accept a callback of the form `(start, stop) -> list[dict]`
- it's not THAT hard for us to be helpful and accept the types that 95% of users will be working with natively.

But I also notice that solara doesn't depend on pandas by itself. So if someone is just making web requests and getting JSON back, it would be nice if they could plot that in a DataTable as a list[dict] without needing to install pandas for a useless round trip.

[...]

PR #600, opened 2024-04-11
refactor API for workign with external dataframe libs

NickCrews:
working towards https://github.com/widgetti/solara/issues/599

### DIFF
--- a/solara/components/datatable.py
+++ b/solara/components/datatable.py
@@ -11,7 +11,7 @@ import solara.hooks.dataframe
 import solara.lab
 import traitlets
 from solara.lab.hooks.dataframe import use_df_column_names
-from solara.lab.utils.dataframe import df_records
+from solara.lab.utils.dataframe import df_len, df_records, df_slice
 
 from .. import CellAction, ColumnAction
 
@@ -89,7 +89,7 @@ def DataTable(
     on_column_header_hover: Optional[Callable[[Optional[str]], None]] = None,
     column_header_info: Optional[solara.Element] = None,
 ):
-    total_length = len(df)
+    total_length = df_len(df)
     options = {"descending": False, "page": page + 1, "itemsPerPage": items_per_page, "sortBy": [], "totalItems": total_length}
     options, set_options = solara.use_state(options, key="options")
     format = format or format_default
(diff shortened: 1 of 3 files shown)


maartenbreddels (changes requested):
Looks good, some minor issues with older Python versions

NickCrews:
Man, do you really still have that many 3.6 users?? Let's abandon them soon, they gotta keep up, catering to them isn't worth the fuss!

maartenbreddels:
Not many, but it's a small effort high reward (we had a large client that still needed to support it)

maartenbreddels:
Btw, please do not merge master back into a branch, but rebase instead.
(Note to self: write this into our contributing document)

NickCrews:
thanks, sorry. Just blindly hit the default button in the github UI. I wish the default were rebase. Perhaps you can change that default in the repo settings?

maartenbreddels:
Doesn't seem to be an option.

maartenbreddels:
Code wise I 💯 with this, but there are actually many changes in this PR, and although I think you did a good job of separating it into different commits, but I'd like to ask for a few changes. I'd ideally see 4 commits.
 * refactor: move df_records() from DataTable to dataframe utils
 * refactor: move slicing of dataframe from DataTable to dataframe utils
 * fix: use .iloc for pandas dataframe slicing
 * refactor: move df_coloumn into dataframe utils

The type-fix commit should move to the commits they belong to, and the merge commit should not be there.

Let me know if you are comfortable with this kinda of rebasing, if not, let me know, and we can try to take it over.

Thanks!

NickCrews:
OK, thanks for the review, I agree that is better. FYI I find that rebasing tedious and disincentivizes me from wanting to contribute again, in my own code I wouldn't bother being as scrupulous as this, I would prioritize a little more dev velocity over a really clean git history. You ultimately are the one who has to live with this code so it's your call of course and I respect that, but FYI I would love to avoid rebasing and merge conflicts as much as possible.

maartenbreddels:
Yes, I understand the tradeoff. We've been hit by a few very difficult to debug bugs in the past that were only possibly to solve with fine commits, so we have a tendency to be a bit strict. Also, it's good to clearly think what the intent is of individual commits, but it can be tedious at times (that's why I suggested I could take over).

In any case, thank you 🙏

maartenbreddels:
Btw, I never merge, because I have no idea what it does internally, and I also have no idea how to interpret the git history after a merge back and forward. By having only a linear history in master, and always rebase again that, I have a good mental model of the code.

PR #624, opened 2024-04-30
fix[docs]: wide outputs of jupyter notebook based tutorials overflowing width limit"

iisakkirotko:

PR #625, opened 2024-04-30
fix!: solara.Meta elements insert empty span into DOM

### LINKED ISSUE #626: `solara.Meta` adds an empty span to DOM
This can have unexpected consequences for spacing when, for instance, using flex-boxes with a set gap, like the ones that are automatically added if an element has more than one direct child.

An example issue can be found in the linked PR

iisakkirotko:
The insertion of `<span></span>` into the dom is what causes strange extra spacing at the top of notebook-based tutorials with solara.Meta tags currently:

image

Fixes #626 

TODO:
- [ ] Revert change to jupyter dashboard tutorial page

### DIFF
--- a/solara/components/head_tag.vue
+++ b/solara/components/head_tag.vue
@@ -1,4 +1,5 @@
-<template><span></span></template>
+<template>
+</template>
 
 <script>
 module.exports = {
--- a/solara/website/pages/documentation/getting_started/content/04-tutorials/60-jupyter-dashboard-part1.py
+++ b/solara/website/pages/documentation/getting_started/content/04-tutorials/60-jupyter-dashboard-part1.py
@@ -10,56 +10,54 @@ title = "Jupyter Dashboard (1/3)"
 
 @solara.component
 def Page():
-    with solara.Column(gap=0):
-        title = "Build your Jupyter dashboard using Solara"
-        solara.Meta(property="og:title", content=title)
-        solara.Meta(name="twitter:title", content=title)
-        solara.Title(title)
+    title = "Build your Jupyter dashboard using Solara"
+    solara.Meta(property="og:title", content=title)
+    solara.Meta(name="twitter:title", content=title)
+    solara.Title(title)
 
-        img = "https://dxhl76zpt6fap.cloudfront.net/public/docs/tutorial/jupyter-dashboard1.webp"
-        solara.Meta(name="twitter:image", content=img)
-        solara.Meta(property="og:image", content=img)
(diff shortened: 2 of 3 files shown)


iisakkirotko:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/widgetti/solara/625?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#625** <a href="https://app.graphite.dev/github/pr/widgetti/solara/625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#624** <a href="https://app.graphite.dev/github/pr/widgetti/solara/624?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

mariobuikhuizen:
Why is this a problem? Why is the span even needed?

iisakkirotko:
@mariobuikhuizen This is what causes the extra spacing at the top of notebook-based tutorials with `solara.Meta` tags currently:

image

Granted this happens because these `span`s get inserted into a flex-box with some gaps defined, so not using that could be an alternative solution. The `span`s seem to be necessary because Vue does not like empty templates, those throw two errors saying:

```
[Vue warn]: Error compiling template:
Cannot use <template> as component root element because it may contain multiple nodes.
```
and
```
[Vue warn]: Error compiling template:
Templates should only be responsible for mapping the state to the UI. Avoid placing tags with side-effects in your templates, such as <script>, as they will not be parsed.
```

mariobuikhuizen:
> @mariobuikhuizen This is what causes the extra spacing at the top of notebook-based tutorials with solara.Meta tags

This would be a better title for the commit/PR.

> The spans seem to be necessary because Vue does not like empty templates, those throw two errors

Strange, I never noticed an error with an empty template.

mariobuikhuizen:
Ok, I checked it, I write empty templates like this:
```html
<template>
</template>
```
This produces no errors.

Doing it like this:
```html
<template></template>
```
Does produce the second error you mentioned.

mariobuikhuizen:
It's a bug in ipyvue that emplty template tags on the same line don't work.

iisakkirotko:
Putting the templates on two different lines fixes both errors. What are your thoughts on the PR? In principle this is a breaking change on the UI side (if someone was unknowingly relying on this for spacing, for example), but I think we can sell it as a fix. I could rename it as suggested, but then the scope of the change seems smaller than it might be.

Issue #611, opened 2024-04-18
PermissionError for nbextensions on startup with v1.29+


brian-c-ogorman:
I'm getting the following error when attempting to run Solara with version 1.29+:

```
PermissionError: [Errno 13] Permission denied: '/usr/local/share/jupyter/nbextensions/jupyter-vuetify/extension.js'
```

Everything works fine on 1.28.0 though. I started a [thread](https://discord.com/channels/1106593685241614489/1230612322318553190/1230612322318553190) in the discord questions channel, but it seems like an issue might be more appropriate.

This is the logging output with the stacktrace:

```
2024-04-18 16:04:17,505 [INFO ] [79501] solara.server.app: app-status check: 81d2c0e7-a41d-4c15-a4b5-8d12eca8cf73 app not started
2024-04-18 16:04:17,527 [INFO ] [79501] solara.server.app: Shut down virtual kernel: 81d2c0e7-a41d-4c15-a4b5-8d12eca8cf73
2024-04-18 16:04:17,574 [INFO ] [79501] solara.server.app: Disconnect page df0beeeb-043e-4b84-b676-2e470826b8ba for kernel 81d2c0e7-a41d-4c15-a4b5-8d12eca8cf73
2024-04-18 16:04:17,576 [INFO ] [79501] solara.server.app: Scheduling kernel cull, will wait for max 86400.0 before shutting down the virtual kernel 81d2c0e7-a41d-4c15-a4b5-8d12eca8cf73
2024-04-18 16:04:20,734 [ERROR] [79501] uvicorn.error: Exception in ASGI application
Traceback (most recent call last):
  File "/.../venv/solara-1-29/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/.../venv/solara-1-29/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
    return await self.app(scope, receive, send)
  File "/.../venv/solara-1-29/lib/python3.10/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/.../venv/solara-1-29/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/.../venv/solara-1-29/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
[...]

PR #648, opened 2024-05-13
fix: catch PermissionError from checking if directory exists

brian-c-ogorman:
Fix for https://github.com/widgetti/solara/issues/611

# Overview

Calls to `jupyter_core.paths.jupyter_path("nbextensions")` (see [here](https://github.com/widgetti/solara/blob/bb109068e51a0434eff88e36e67f889cad36cf94/solara/server/server.py#L428)) may return directories that do not exist. However, the user may not have permission to read some of these paths, leading to a `PermissionError` when calling `pathlib.Path.exists()` (see [here](https://github.com/widgetti/solara/blob/bb109068e51a0434eff88e36e67f889cad36cf94/solara/server/server.py#L434) and [here](https://github.com/widgetti/solara/blob/bb109068e51a0434eff88e36e67f889cad36cf94/solara/server/server.py#L447)). The `PermissionError` ends up crashing the application.

## Fix

This change catches the `PermissionError` and logs it as a warning, preventing the application from crashing. Essentially: if the user does not have permission to the path, it might as well not exist.

### DIFF
--- a/solara/server/server.py
+++ b/solara/server/server.py
@@ -431,8 +431,12 @@ def get_nbextensions() -> Tuple[List[str], Dict[str, Optional[str]]]:
 
     def exists(name):
         for directory in nbextensions_directories:
-            if (directory / (name + ".js")).exists():
-                return True
+            try:
+                file_path = directory / (name + ".js")
+                if file_path.exists():
+                    return True
+            except PermissionError:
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thank you, Brian, clean code and clean commit message 👍

PR #657, opened 2024-05-21
refactor!: streamline and improve applayout api

### LINKED ISSUE #584: creating a multi-page app without forcing an appbar
Is it possible to create a multi-page app without forcing the creation of an app bar with tabbed pages?  I effectively have two "single-page" apps, each with their own top-level `Page` component,  that I need to combine into a multi-page app, with each behind a route.  I tried something like the following:

```
import solara

from sdss_solara.pages.jdaviz_embed import Page as Embed
from sdss_explorer.pages import Page as Dashboard

@solara.component
def Home():
    solara.Markdown("Solara Home")


routes = [
    solara.Route(path="/", component=Home, label="Home"),
    solara.Route(path="embed", component=Embed, label="Embed"),
    solara.Route(path="dashboard", component=Dashboard, label="Dashboard"),
]
```
but this forces an AppBar, which covers up components on the "embed" page.   Additionally, navigating to the home page first correctly displays the "Solara Home" text, but then toggling to the embed page and back to home, the content missing.  

In this case I don't need the page navigation, I just need to point to "site/solara/embed" and "site/solara/dashboard" separately and have them render as expected.

Also for reference, I am not running this with the solara server, the combined app will be sub-mounted into an existing FastAPI app. I think that just means that the `SOLARA_APP` env variable should point to the top-level solara application module.  

I tried disabling the layout by adding the following in the above top-level module but that didn't seem to change anything. 
```
@solara.component
def Layout(children=[]):
    # there will only be 1 child, which is the Page()
    return children[0]
```

iisakkirotko:
Fixes #584.

Includes four breaking changes:
- A change in the default color of the `AppBar` from `"primary"` to the Vuetify default (light/dark grey depending on theme)
- Changes the behaviour of the `navigation` argument, where previously it would hide the entire `AppBar` it now correctly only hides the tab navigation (if present)
- Changes the default behaviour of `show_app_bar` (which is now also available as an argument) to where it no longer requires the `title`-argument to be set for `show_app_bar` to be true.
- Do not turn the first non-`AppBar` child of `AppLayout` into a sidebar element.

### DIFF
--- a/solara/components/applayout.py
+++ b/solara/components/applayout.py
@@ -196,9 +196,10 @@ def AppLayout(
     children=[],
     sidebar_open=True,
     title=None,
+    show_app_bar: Optional[bool] = None,
     navigation=True,
-    toolbar_dark=True,
-    color: Optional[str] = "primary",
+    toolbar_dark: Optional[bool] = None,
+    color: Optional[str] = None,
     classes: List[str] = [],
     style: Optional[Union[str, Dict[str, str]]] = None,
 ):
@@ -222,9 +223,14 @@ def AppLayout(
 
     # Arguments
 
-     * `children`: The children of the AppLayout. The first child is used as the sidebar content, the rest as the main content.
+     * `children`: The children of the AppLayout. Note: Since Solara 2.0, the first child is no longer placed in the sidebar.
+        Use [Sidebar](/documentation/components/layout/sidebar) instead.
      * `sidebar_open`: Whether the sidebar is open or not.
      * `title`: The title of the app shown in the app bar, can also be set using the [Title](/documentation/components/page/title) component.
+     * `show_app_bar`: Whether the app bar should be shown. If `None` (the default), `AppBar` is shown if:
+         * There are one or more sibling routes to the current page.
+         * **OR**: There are one or more children of the `AppBar` component.
+         * **OR**: There are one or more children of the `AppBarTitle` component.
      * `toolbar_dark`: Whether the toolbar should be dark or not.
      * `navigation`: Whether the navigation tabs based on routing should be shown.
      * `color`: The color of the toolbar.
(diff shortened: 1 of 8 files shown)


iisakkirotko:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/widgetti/solara/657?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#657** <a href="https://app.graphite.dev/github/pr/widgetti/solara/657?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#656** <a href="https://app.graphite.dev/github/pr/widgetti/solara/656?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

iisakkirotko:
As said in https://github.com/widgetti/solara/pull/1003#issuecomment-2836334610, this also fixes https://github.com/widgetti/solara/issues/1002.

maartenbreddels (approved):
Nice. PR does include some extra commits, I guess that's a graphite issue?

solara/components/applayout.py
```diff
@@ -282,7 +281,7 @@ def set_path(index):
     if (tabs_element is None) and routes and navigation and (len(routes) > 1):
         with solara.lab.Tabs(value=index, on_value=set_path, align="center") as tabs_element:
             for route in routes:
-                name = route.path if route.path != "/" else "Home"
+                name = route.label if route.label is not None else (route.path if route.path != "/" else "Home")
```
johannesWen, line 284: I need a custom route.label for the "/" route. Any change to see this in the current release?
@iisakkirotko @maartenbreddels

PR #658, opened 2024-05-24
bypass switch color to vuetify

johannesWen:
Bypass the color property of the solara.Switch component to the vuetify Switch.

### DIFF
--- a/solara/components/switch.py
+++ b/solara/components/switch.py
@@ -12,6 +12,7 @@ def Switch(
     value: Union[bool, solara.Reactive[bool]] = True,
     on_value: Callable[[bool], None] = None,
     disabled: bool = False,
+    color: str = None,
     children: list = [],
     classes: List[str] = [],
     style: Optional[Union[str, Dict[str, str]]] = None,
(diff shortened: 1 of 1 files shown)


solara/components/switch.py
```diff
@@ -46,6 +47,7 @@ def Page():
      * `value`: The current value of the switch (True or False).
      * `on_value`: A callback that is called when the switch is toggled.
      * `disabled`: If True, the switch is disabled and cannot be used.
+     * `color`: The color of the switch. Can be any of the Vuetify color classes.
```
iisakkirotko, line 50:
```suggestion
     * `color`: The color of the switch. Can be the name of a [theme](https://solara.dev/documentation/components/lab/theming) color variable (i.e. "success"), a color from the [Material color palette](https://v2.vuetifyjs.com/en/styles/colors/#material-colors) (i.e. "red darken-3"), or a CSS color (i.e. "#ff991f").
```

I think it's best to elaborate here, since Solara users might not be familiar with Vuetify.

iisakkirotko (changes requested):

solara/components/switch.py
johannesWen: Good point, thank you.

iisakkirotko (approved):
Looks good to me. Thanks a lot for the PR!

maartenbreddels:
Thank you both!

PR #667, opened 2024-06-04
docs: todo example to use a class for state

kecnry:
This PR updates the todo example to place all state in a (re-usable) `TodoItem` class, instead of a `dataclass`.  This allows avoiding the need for `Ref` and `.fields`.

If there is no other example for `Ref` and `fields` in the docs, then maybe this shouldn't be merged, but at least shows the difference between the two implementations.

### DIFF
--- a/solara/website/pages/documentation/examples/utilities/todo.py
+++ b/solara/website/pages/documentation/examples/utilities/todo.py
@@ -22,7 +22,6 @@ from typing import Callable
 import reacton.ipyvuetify as v
 
 import solara
-from solara.lab.toestand import Ref
 
 
 class TodoItem:
--- a/solara/website/pages/documentation/examples/utilities/todo.py
+++ b/solara/website/pages/documentation/examples/utilities/todo.py
@@ -1,65 +1,22 @@
 """# Todo application
 
-Demonstrates the use of reactive variables in combinations with dataclasses.
-
-With solara we can get a type safe view onto a field in a dataclass, pydantic model, or
-attr object.
-
-This is using the experimental `solara.lab.Ref` class, which is not yet part of the
-official API.
-
+Demonstrates the use of reactive variables in an externally defined state class.
 
 ```python
-import dataclasses
 import solara
-from solara.lab import Ref
 
-@dataclasses.dataclass(frozen=True)
 class TodoItem:
-    text: str
-    done: bool
+    def __init__(self, text='', done=False):
+        self.text = solara.reactive(text)
+        self.done = solara.reactive(done)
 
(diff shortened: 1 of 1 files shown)


solara/website/pages/documentation/examples/utilities/todo.py
```diff
@@ -141,7 +97,7 @@ def create_new_item(*ignore_args):
 ]
 
 
-# We store out reactive state, and our logic in a class for organization
+# We store our reactive state, and our logic in a class for organization
 # purposes, but this is not required.
 # Note that all the above components do not refer to this class, but only
```
kecnry, line 102: these in-line comments may need updating?

solara/website/pages/documentation/examples/utilities/todo.py
```diff
@@ -141,7 +96,7 @@ def create_new_item(*ignore_args):
...
 # to do the Todo items.
```
hsunner, line 102: This could use further justification and explanation.

solara/website/pages/documentation/examples/utilities/todo.py
```diff
@@ -151,25 +106,23 @@ def create_new_item(*ignore_args):
```
hsunner, line 106: If this is an example of how to manage state, it would be even more valuable to show _how_ the todos can be reused. Now the Page component is effectively a Todo component, but it's not reusable as State can only manage a single list of Todos.

solara/website/pages/documentation/examples/utilities/todo.py
```diff
@@ -183,14 +136,16 @@ def TodoStatus():
         solara.Success("All done, awesome!", dense=True)
 
 
+state = State(initial_items)
+
+
 @solara.component
 def Page():
     with solara.Card("Todo list", style="min-width: 500px"):
```
hsunner, line 144: Please consider doing a separate Todo component card and show how it can be reused. Again, this may be another example (a dashboard of TODO lists for different people/tasks where you can add or remove entire TODOs?)

solara/website/pages/documentation/examples/utilities/todo.py
```diff
@@ -183,14 +136,16 @@ def TodoStatus():
         solara.Success("All done, awesome!", dense=True)
 
 
+state = State(initial_items)
```
hsunner, line 139: It would be nice with a comment explaining how the state object can still work for multiple clients (virtual kernels), despite it not being declared a solara.reactive.

hsunner (commented):
There seem to be a bug in the TodoEdit class. It will update the edited TodoItem even when "close" is pressed.
Also including some comments on how the example could be made more enlightening (at least to me).

PR #678, opened 2024-06-10
feat: add mhchem katex extension

Jhsmit:
mhchem package provides commands for typesetting chemical molecular formulae and equations

mhchem katex extension docs are [here](https://github.com/KaTeX/KaTeX/blob/main/contrib/mhchem/README.md)

this contribution is a combination of copy-paste with suggestions from codeium

example:

```python
import solara

md_text = """
Normal math:

$$
1 + 1 = \sqrt{2}
$$

mhchem:

$$
\ce{6CO2 + 6H2O -> C6H12O6 + 6O2}
$$

"""

@solara.component
def Page():
    solara.Markdown(md_text)
```

output:
image

### DIFF
--- a/solara/components/markdown.py
+++ b/solara/components/markdown.py
@@ -92,6 +92,7 @@ module.exports = {
         await this.loadRequire();
         this.mermaid = await this.loadMermaid();
         this.mermaid.init();
+        await this.loadMHchem();
         this.latexSettings = {
                 delimiters: [
                     {left: "$$", right: "$$", display: true},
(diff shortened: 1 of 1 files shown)


solara/components/markdown.py
```diff
@@ -157,6 +158,11 @@ def _markdown_template(
         async loadMermaid() {
             return (await this.import([`${this.getCdn()}/mermaid@10.8.0/dist/mermaid.min.js`]))[0]
         },
+        async loadMHchem() {
+            await this.loadKatex();
```
maartenbreddels, line 162:
```suggestion
            await this.loadKatex();
```
```suggestion
```

This will cause the CSS to be added twice,

solara/components/markdown.py
```diff
@@ -92,6 +92,7 @@ def _markdown_template(
         await this.loadRequire();
         this.mermaid = await this.loadMermaid();
         this.mermaid.init();
+        await this.loadMHchem();
```
maartenbreddels, line 95: I guess we need to move that to after line 110, if/when we use katex.

maartenbreddels (changes requested):

Jhsmit:
I've now added the loading of mhchem to the same `this.import` call which loads katex. It seems like the better solution.

PR #693, opened 2024-06-24
Pdf preview component

ntjess:
**Fast** PDF previewing using solara 🤝 pdf.js. I adapted the example [here](https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.mjs) to work as a vue component.

In the future, we can enable the search bar (commented out currently), annotations, etc. But this may be a good start.

I'm open to any feedback! Question for solara devs: I am currently passing PDF bytes to the backend, is it possible to send the file directly? We can also expand to allow URLs easily, since pdfjs natively supports them as a data source.

Example loading a 3000 page PDF as a b64-encoded string:
```python
import solara
import requests
import base64


@solara.component
def Page():
    # Test with a 3000 page pdf! (45MB)
    pdf = requests.get("https://matplotlib.org/2.0.2/Matplotlib.pdf").content
    name_b64_map = {"matplotlib.pdf": base64.b64encode(pdf).decode()}
    current_file = solara.use_reactive("")
    with solara.Column(style={"width": "800px"}):
        solara.PDFViewer(pdf_name_b64_map=name_b64_map, current_file=current_file)
        solara.Button("Load!", on_click=lambda: current_file.set("matplotlib.pdf"))

        solara.Button("Unload", on_click=lambda: current_file.set(""))
```

### DIFF
--- a/solara/components/__init__.py
+++ b/solara/components/__init__.py
@@ -15,6 +15,7 @@ from .details import Details  # noqa: F401 F403
 from .file_browser import FileBrowser  # noqa: F401 F403
 from .image import Image  # noqa: F401 F403
 from .markdown import Markdown, MarkdownIt  # noqa: F401 F403
+from .pdf_viewer import PDFViewer  # noqa: F401 F403
 from .slider import (  # noqa: F401 F403
     DateSlider,  # noqa: F401 F403
     FloatSlider,  # noqa: F401 F403
--- a/solara/components/pdf_viewer.py
+++ b/solara/components/pdf_viewer.py
@@ -1,0 +1,30 @@
+from typing import cast
+import ipyvue
+import solara
+import traitlets
+
+
+class PDFViewerWidget(ipyvue.VueTemplate):
+    template_file = (__file__, "pdf_viewer.vue")
+
+    pdf_name_b64_map = traitlets.Dict({}).tag(sync=True)
+    current_file = traitlets.Unicode("").tag(sync=True)
+    height = traitlets.Unicode("500px").tag(sync=True)
+
+    def __repr__(self):
+        return f"PDFViewerWidget(files={list(self.pdf_name_b64_map)}, active_file={self.current_file})"
+
+
+@solara.component
+def PDFViewer(
+    pdf_name_b64_map: dict[str, str], current_file: solara.Reactive[str], **kwargs
+):
+    if pdf_name_b64_map is None:
+        pdf_name_b64_map = {}
+
+    def on_load():
+        real = cast(PDFViewerWidget, solara.get_widget(widget))
+        real.current_file = current_file.value
+
+    solara.use_effect(on_load, [current_file.value])
+    widget = PDFViewerWidget.element(pdf_name_b64_map=pdf_name_b64_map, **kwargs)  # type: ignore
--- a/solara/components/pdf_viewer.vue
+++ b/solara/components/pdf_viewer.vue
@@ -1,3 +1,4 @@
+// Credit for js example: https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.mjs
 <template :pdf_name_b64_map="pdf_name_b64_map" :current_file="current_file">
   <div id="pdfViewerRoot" :style="'height: ' + height">
     <div id="viewerContainer">
@@ -68,8 +69,6 @@ module.exports = {
 
       // Some PDFs need external cmaps.
       //
-      const DEFAULT_URL =
-        "https://github.com/mozilla/pdf.js/raw/master/web/compressed.tracemonkey-pldi-09.pdf";
       // const CMAP_URL = "../../node_modules/pdfjs-dist/cmaps/";
       const CMAP_PACKED = true;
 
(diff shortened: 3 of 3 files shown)


maartenbreddels:
Awesome work!

Do you think it makes sense to have this in solara as a component, or would it make sense to be its own package, e.g. `solara-pdf-viewer`? 

I'll get back with some more feedback, and best practices, so we can start polishing here first.

ntjess:
I didn't see a precedent for independent solara packages, otherwise that would make sense too. Nonetheless, I would push for first-party integration within solara until there is a guide describing:
- [ ] How to make third-party extensions offline compatible (i.e., how could we make `solara-pdf-viewer` compatible with `solara-assets` stuff?)
- [ ] General recommended practices for standalone components (e.g. a template repo). Currently, the only third-party packages referenced as examples under https://solara.dev/docs are full apps, not reusable components.
- [ ] More firm calling conventions (i.e. does every component need to accept `on_value`, `v_on_model`, ...)? It is a recipe for incompatibility later on as-is.
    - It is easier to 'police' how these components and their API are structured when solara authors can reject the PR

I'm curious of your thoughts as well!

maartenbreddels:
I am thinking this is a good motivation to get this document and/or start a precedent. This can serve as an example. What do you think?

ntjess:
That works for me! A few questions/requests for you to get started:
- [ ] Create a `solara-components` (or equivalent) GH repo owned by solara devs
    - As a solara user, I would want to know which non-builtin widgets are "blessed" by solara devs (offline compatible, conform to a specific API, etc.) so I can gauge how much risk is involved with using the third-party widget
        - This is especially relevant for the first few external components. Solara is new enough that as a user, I would find comfort that third-party options under a solara-owned repo take into considerations the edge cases of vue template initialization, nonblocking threads/tasks where applicable, efficiency of loading, etc. These factors would provide an estimate of the widget's reliability.
    - It is more scoped than using the `widgetti` namespace directly
    - Easier for the main solara website to know what to showcase as extensions
- [ ] Create a checklist of the important factors a widget needs to fulfill to become "blessed". If this checklist exists, third-party widgets can at least specify where they are compliant even if not all attributes are implemented. Off the top of my head:
    - [ ] Conformant API:
        - `value` can be reactive or non-reactive, `on_value` is an optional callable
        - Use `values` instead of `options` for selection-like parameters, etc.
        - Use `snake_case` python-facing variables in vue templates
        - Extra `kwargs` can be passed to the underlying vue element (Note: this is not currently the case for `FileDrop` but that's a separate request. I'd like to pass an `accept` list to the solara component but it's not currently possible)
        - Recommendations on `VueTemplate` with `template_file` vs. `template` vs. `solara.component_vue(...)`, `ipyvuetify` vs. `vuetify` as a base
    - [ ] Awareness of dark/light mode (and potentially other theme variables like `primary`)
        - We should standardize the way style is applied to the elements. Do you recommend an inline `solara.Style(...)` like the chat elements? `style="..."` on each element? `classes=[...]` with a separate `style.css` sheet the user can optionally apply?
    - [ ] Offline mode supported
    - [ ] At least one test case / example file that can be linked by the main website at a later time
    - [ ] Every parameter documented
        - Solara doesn't use a standard numpydoc/googledoc/rst format, so perhaps a "docs guide" for how to conform to your markdown sections is useful. (Or, solara can migrate to a more standard format 😛 )

maartenbreddels:
What about we start documenting what you need to know in order to do this, and we'll collaborate on those docs. As part of that I'd like to also have a cookie-cutter, so it's easier for people to repeat.

We can 'bless' (love that choice of word) your pdf viewer by listing it on our docs, but I think having a repo in your own org/name would make sense first.

Regarding docs, and docstrings format: We are actively thinking about the current website, our approach, and if we should change that due to the existence of https://py.cafe (we can host example there, instead of running it in the solara server directly). My preference goes to an mkdocs based docs system. Curious where your preference lies.

I can start writing docs and a cookiecutter/template next week to answer your questions in two forms (writing and code). Does that sound like a good first plan?

ntjess:
This sounds great to me, thanks.

Regarding docs style: I'm preferential to NumpyDoc, since it formats nicely enough as markdown even without a docstring-aware parser (compared to e.g. Google-style). It is also widely supported, and tools like `docstring-parser` can consume it for programmatic use. It is also less verbose and much more source-readable than rst-style.

I haven't used mkdocs before, but it seems, if desired, there is at least an extension to handle most standard docstring styles (including numpy): https://mkdocstrings.github.io/python/#features.

> due to the existence of https://py.cafe/

The playground is a great idea in theory! It would allow users to quickly test out modifications to the sample code. **However**, I love the immediate load time of examples in the current solara docs page. In contrast, it often takes 30+ seconds for py.cafe to load for me, and this would become a bit frustrating if the wait time was required for every interactive example... Would you be providing links to a "pre-warmed" instance?

> We are actively thinking about the current website, our approach...
The website is a great starting point as-is! If you are open to feedback, my biggest challenges with the current configuration are:
- It doesn't work on mobile -- the pages freeze, left and right halves of pages get cut off, search and navigation often don't work, ...
- (minor nitpick) Search results obscure what you are typing when and expand above the search bar. It makes erasing previous text difficult when searching adaptively
    - I often find it easier to navigate directly to the repo folder and use GH search
- Limited documentation for `ipyvue` and `ipyvuetify` components:
    -  i.e., `v_model` is often typed as `Any`, without a clear indication of what the type will resolve to
    -  It's unclear when you must provide a `v_model`
    - No docstrings on parameters (but this is a much bigger ask and can wait)

Again, these are mostly minor issues! The code is very understandable and there are enough examples to gather a comprehensive understanding of how the pieces fit together.

I'll close this PR for now and await updates regarding the "extension" template.

PR #700, opened 2024-07-03
Fix: docs navigation

maartenbreddels:
We did not have a node in the sidebar for the quickstart, which made me realized we were missing some pages.

### DIFF
--- a/solara/website/components/sidebar.py
+++ b/solara/website/components/sidebar.py
@@ -73,7 +73,7 @@ def Sidebar():
                                         with solara.Link(
                                             path,
                                         ):
-                                            with solara.v.ListItem(dense=True, style_="padding: 0 20px;", value=path):
+                                            with solara.v.ListItem(dense=True, style_="margin-left: 40px; padding: 0 20px;", value=path):
                                                 solara.v.ListItemContent(
                                                     children=[subitem.label],
                                                 )
--- a/solara/website/components/sidebar.py
+++ b/solara/website/components/sidebar.py
@@ -24,6 +24,8 @@ def Sidebar():
                                 solara.v.ListItemIcon(children=[solara.v.Icon(children=["mdi-home"])])
                             solara.v.ListItemTitle(style_="padding: 0 20px;", children=[route.label])
                 else:
+                    path_top_level = "/documentation/" + route.path
+                    top_level_expanded = router.path.startswith(path_top_level)
                     with solara.v.ListGroup(
                         v_slots=[
                             {
@@ -34,11 +36,13 @@ def Sidebar():
                                 ),
                             }
                         ],
-                        value=router.path.startswith("/documentation/" + route.path),
+                        value=top_level_expanded,
                     ):
                         for item in route.children:
                             if item.path == "/":
                                 continue
+                            path_sub = "/documentation/" + route.path + "/" + item.path
+                            sub_should_be_expanded = router.path.startswith(path_sub)
                             if item.children != [] and any([c.label is not None and c.path != "/" for c in item.children]):
                                 with solara.v.ListGroup(
                                     v_slots=[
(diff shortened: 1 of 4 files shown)


solara/website/components/sidebar.py
```diff
@@ -34,58 +37,61 @@ def Sidebar():
...
+                                            children=[label],
+                                        )
+        with solara.Link("/contact"):
+            with solara.v.ListItem():
+                solara.v.ListItemIcon(children=[solara.v.Icon(children=["mdi-email"])])
+                solara.v.ListItemTitle(style_="padding: 0 20px;", children=["Contact"])
+        with solara.Link("/changelog"):
+            with solara.v.ListItem():
+                solara.v.ListItemIcon(children=[solara.v.Icon(children=["mdi-history"])])
+                solara.v.ListItemTitle(style_="padding: 0 20px;", children=["Changelog"])
```
iisakkirotko, line 95:
```suggestion
            with solara.Link("/contact"):
                with solara.v.ListItem(value="/contact"):
                    solara.v.ListItemIcon(children=[solara.v.Icon(children=["mdi-email"])])
                    solara.v.ListItemTitle(style_="padding: 0 20px;", children=["Contact"])
            with solara.Link("/changelog"):
                with solara.v.ListItem(value="/changelog"):
                    solara.v.ListItemIcon(children=[solara.v.Icon(children=["mdi-history"])])
                    solara.v.ListItemTitle(style_="padding: 0 20px;", children=["Changelog"])
```

Because these items are now no longer in the `ListItemGroup`, they are not highlighted when you're viewing them. You also don't get the background colour change when hovering over them.

solara/website/components/sidebar.py
```diff
@@ -34,58 +37,61 @@ def Sidebar():
...
                                     no_action=True,
-                                    value=router.path.startswith("/documentation/" + route.path + "/" + item.path),
+                                    eager=True,  # better for SEO
+                                    value=sub_should_be_expanded,
                                 ):
                                     for subitem in item.children:
-                                        # skip pages that are only used to demonstrate Link or Router usage
-                                        if subitem.path == "/" or subitem.label is None:
+                                        # skip the 'homepage' of the examples only
+                                        if subitem.path == "/" and route.path == "examples":
```
iisakkirotko, line 67:
```suggestion
                                        if subitem.path == "/" and route.path != "getting_started":
```

This causes for a link to an invalid route ("Not supposed to be rendered") to be present under each subsection of the components and api sections, so those should at least be excluded.

Also, a lot of the subsection overview pages under advanced are also very half baked (see [here](https://solara.dev/documentation/advanced/development/) for an example), so I think we should either improve those, or exclude them as well (as I've done in the suggestion).

iisakkirotko (changes requested):
A lot of good improvements, I already noticed that only one item was highlighted much more consistently in the menu now.

solara/website/components/sidebar.py
iisakkirotko: Seeing as we want to add content to the "enterprise" overview page, I suppose we can improve the content of the other advanced overview pages in that PR and merge at the same time?

solara/website/components/sidebar.py
```diff
@@ -34,58 +37,61 @@ def Sidebar():
...
                                             if item.path != "fullscreen"
                                             else "/apps/" + subitem.path
                                         )
                                         with solara.Link(
                                             path,
                                         ):
-                                            with solara.v.ListItem(dense=True, style_="padding: 0 20px;", value=path):
+                                            with solara.v.ListItem(dense=True, style_="padding-left: 60px; padding-right: 20px", value=path):
                                                 solara.v.ListItemContent(
                                                     children=[subitem.label],
```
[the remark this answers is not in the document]
maartenbreddels: doesn't happen in practice so I'm ok with it, but sharp observation

PR #709, opened 2024-07-10
fix: fix pywebview compatibility by using requirejs correctly

maartenbreddels:
It seems we cannot (what is now called loadApp) outside a requirejs call, with the jupyter-vue dependencies. Otherwise this results in pywebview seeing jupyter-vue as null. Not sure why this does not happen with other browsers.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -217,6 +217,94 @@ jobs:
         run:
           pip install packages/solara-meta/dist/*.whl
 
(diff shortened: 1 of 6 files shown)


maartenbreddels:
qt seems like a better, more support solution. The dependencies of pywebview do not seem to be maintained that well, and it does not seem stable in CI (also not on pywebview's CI)

Instead, see https://github.com/widgetti/solara/pull/724

PR #712, opened 2024-07-11
ci: did not run windows integration tests

maartenbreddels:


maartenbreddels:
debugging windows will take some time, we'll first merge this, most obvious issues are solved now

Issue #714, opened 2024-07-15
Bug: solara server hits Internal Server Error due to get_nbextensions logic


sjdemartini:
(Solara looks awesome; thank you for building it! I'm just trying it out for the first time.)

Copying the example [ipywidgets code from the Solara docs](https://solara.dev/documentation/getting_started/tutorials/ipywidgets) and putting it into a `solara-ipywidgets-test.ipynb` file, then running the Solara server, it results in an error in solara's code for nbextensions. This shows `Internal Server Error` in the browser and the following stack trace in the shell:

```
❯ solara run solara-ipywidgets-test.ipynb:button
Solara server is starting at http://localhost:8765
I get run at startup, and for every page request
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/path/to/env/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 396, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/path/to/env/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 70, in __call__
    return await self.app(scope, receive, send)
  File "/path/to/env/lib/python3.10/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/path/to/env/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/path/to/env/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/path/to/env/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/path/to/env/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/path/to/env/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/path/to/env/lib/python3.10/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/path/to/env/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/path/to/env/lib/python3.10/site-packages/starlette/routing.py", line 756, in __call__
[...]

PR #715, opened 2024-07-15
Fix bug in server.py glob iteration by skipping directories

sjdemartini:
Fixes https://github.com/widgetti/solara/issues/714

### DIFF
--- a/solara/server/server.py
+++ b/solara/server/server.py
@@ -452,7 +452,7 @@ def get_nbextensions() -> Tuple[List[str], Dict[str, Optional[str]]]:
                 file_path = directory / (name + ".js")
                 if file_path.exists():
                     for file in directory.glob("**/*.*"):
-                        if file.is_file():  # Skip directories
+                        if file.is_file():  # Otherwise directories with a dot in the name are included
                             data = file.read_bytes()
                             h.update(data)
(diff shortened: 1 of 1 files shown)


solara/server/server.py
```diff
@@ -452,8 +452,9 @@ def hash_extension(name):
                 file_path = directory / (name + ".js")
                 if file_path.exists():
                     for file in directory.glob("**/*.*"):
-                        data = file.read_bytes()
-                        h.update(data)
+                        if file.is_file():  # Skip directories
```
iisakkirotko, line 455:
```suggestion
                        if file.is_file():  # Otherwise directories with a dot in the name are included
```

I think it would be a good idea to indicate in the comment why we need to explicitly skip directories, since one might expect that the above pattern doesn't include any.

iisakkirotko (commented):
Thanks a lot for the PR @sjdemartini! @maartenbreddels I think we can include this when you do the next release.

dennisjlee:
Could we merge this please? I also ran into this issue and it's blocking me from trying Solara.

maartenbreddels:
We'll do a release very soon, thanks for the fix @sjdemartini, and for the review @iisakkirotko

PR #727, opened 2024-08-05
Fix typo in 10-oauth.md

ktaletsk:


iisakkirotko (approved):
Thanks a lot for the PR! The remaining failure is unrelated, and should be fixed by #729.

PR #742, opened 2024-08-21
pre-commit: Add auto-update schedule

EwoutH:
To make sure the pre-commit configuration stays up to date, add auto-update schedule to the pre-commit configuration file.

[pre-commit.ci](https://pre-commit.ci/) needs to be enable for this (which I highly recommend). When enabled, it will check once a week if any pre-commit action can be updated, and will open a PR if so.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+ci:
+    autoupdate_schedule: 'weekly'
+
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0


maartenbreddels:
Not so sure if I will like it in practice, but I've added it, and if I change my mind we can remove it.

Thanks!

maartenbreddels (approved):

maartenbreddels:
I think I did, I enabled this repo after logging in, maybe I missed
something?

EwoutH:
Perfect, thanks for confirming!

PR #749, opened 2024-08-23
fix: allow embedding in iframe that restrict local storage

maartenbreddels:


maartenbreddels:
keeping this open, because I wonder how we can test this.

PR #752, opened 2024-08-26
Require Python 3.8, update CI

EwoutH:
Python 3.6 and 3.7 have both been end-of-life for over a year. This PR starts requiring Python 3.8 and higher for Solara. CI and test code is modified in sync.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.12"
 
       - uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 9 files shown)


EwoutH:
Seems all 3.12 tests get stuck on a similar issue:
```
ImportError while loading conftest '/home/runner/work/solara/solara/tests/conftest.py'.
tests/conftest.py:5: in <module>
    from dotenv import load_dotenv
E   ModuleNotFoundError: No module named 'dotenv'
```

EwoutH:
It seems it just isn't installed when imported here:
https://github.com/widgetti/solara/blob/4647131eea7f05b305a0b0ac3896e37f8111e1e2/tests/conftest.py#L5

Weird that it happens on 3.12 only. But where should I specify it as an required package or install it?

EwoutH:
More complex than I thought, and unfortunately no time to debug it. Closing.

(I would still recommend dropping Python 3.6 and 3.7)

maartenbreddels:
It is weird, python-dotenv *is* a dependency. Might pick up the wrong Python env?

I'd rather extend the Python versions, and only drop the older ones if there are strong reasons (we saw Python 3.6 envs in our consultancy work, and we couldn't do the work if we didn't support it).

Issue #797, opened 2024-09-25
Solara fails to import with Python 3.13 on Windows


EwoutH:
When we were updating our CI setup (https://github.com/projectmesa/mesa/pull/2325) we noticed that the Windows build threw some errors:

```
ERROR tests/test_cell_space.py - AttributeError: module 'solara' has no attribute 'component'.
ERROR tests/test_devs.py - AttributeError: module 'solara' has no attribute 'component'.
ERROR tests/test_solara_viz.py - AttributeError: module 'solara' has no attribute 'reactive'.
!!!!!!!!!!!!!!!!!!! Interrupted: 3 errors during collection !!!!!!!!!!!!!!!!!!!
```

Both Python 3.13 on macOS and Ubuntu and Python 3.12 on Windows run fine, it's only Python 3.13 on Windows that gives the errors.

Since the pip install step takes a lot longer, it could be that either Solara or a dependency of Solara is being build manually on Python 3.13 Windows instead of using a wheel, which could be causing the problems.

The Python 3.13.0 stable release is this Tuesday (2024-10-01), it might be nice if everything runs smoothly on stable release.

EwoutH:
Issue reproduced on Solara's own CI: https://github.com/widgetti/solara/actions/runs/11027367292/job/30625621676?pr=798

![image](https://github.com/user-attachments/assets/c28071d3-d1ea-4b54-a03d-4c16836aab8e)

```
ERROR: Cannot install jupyter-client because these package versions have conflicting dependencies.

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

It seems like a relatively straightforward dependency conflict that just needs to be resolved.

EwoutH:
Hi @maartenbreddels and @iisakkirotko! Python 3.12 stable is releasing in 5 days, any chance these issues can be given some priority? See also:
- https://github.com/widgetti/solara/pull/798

iisakkirotko:
Hi @EwoutH! Unfortunately I'm at a conference for this week, so I won't be able to take a look at this before Monday. I do hope to work on it as soon as I'm back at the office, unless Maarten gets to it before I do.

EwoutH:
Thanks for getting back.

Today is 3.13 release day, so if you have a chance, that would be appreciated!

[...]

PR #798, opened 2024-09-25
CI: Test with Python 3.13

EwoutH:
Update the CI to use Python 3.9 as lowest and Python 3.13 as highest version.

This is mainly a test PR to help you discover if you're ready for Python 3.13, which has its stable release is this Tuesday (2024-10-01).

You can also merge it if you like.

See
- https://github.com/widgetti/solara/issues/797.


maartenbreddels:
Thanks for opening this. I don't understand why our ci continue running even though the pip install fails (with probably a non-zero error code)

PR #805, opened 2024-10-01
feat: render ipypopout content in jupyter notebook and lab

maartenbreddels:
Instead of having to rely on voila to display widgets in a popout window, solara itself now can render it. This allows project to drop the dependency on voila.

### DIFF
--- a/solara/server/jupyter/solara.py
+++ b/solara/server/jupyter/solara.py
@@ -1,0 +1,91 @@
+import json
+import logging
+import os
+from pathlib import Path
+
+import tornado.web
+from jupyter_server.base.handlers import JupyterHandler
+import solara.server.server as server
+
+from solara.server.utils import path_is_child_of
+import solara
+
+logger = logging.getLogger("solara.server.jupyter.solara")
+
+
+@solara.component
+def Page():
+    solara.Error("Hi, you should not see this, we only support ipypopout for now")
+
+
+class SolaraHandler(JupyterHandler):
+    async def get(self, path=None):
+        try:
+            # base url ends with /
+            base_url = self.settings["base_url"]
+            # root_path's do not end with /
+            jupyter_root_path = ""
+            if base_url and base_url.endswith("/"):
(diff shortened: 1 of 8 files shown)


solara/server/jupyter/solara.py
```diff
@@ -0,0 +1,92 @@
...
+            abspath = os.path.abspath(os.path.join(root, path))
+            # return early if someone tries to access a file outside of the directory
+            if os.path.exists(abspath):
+                self.root = root  # make sure all the other methods in the base class know how to find the file
+                break
+
+        # tornado probably already does this, but just to be sure we don't serve files outside of the directory
+        # and that we are consistent with starlette and flask
+        if not path_is_child_of(Path(abspath), Path(self.root)):
+            raise PermissionError(f"Trying to read from outside of cache directory: {abspath} is not a subdir of {self.root}")
```
iisakkirotko, line 75: Wouldn't this still allow someone to find what files exist outside a given root directory? Not sure if that is problematic in this case.

solara/server/jupyter/solara.py
```diff
@@ -0,0 +1,92 @@
...
+        self.roots = paths
+        super().initialize(path=paths[0], default_filename=default_filename)
+
+    def get_absolute_path(self, root: str, path: str) -> str:  # type: ignore
+        # find the first absolute path that exists
+        self.root = self.roots[0]
+        abspath = os.path.abspath(os.path.join(root, path))
+        for root in self.roots[1:]:
+            abspath = os.path.abspath(os.path.join(root, path))
+            # return early if someone tries to access a file outside of the directory
```
iisakkirotko, line 67: I don't think I follow what this comment is trying to say.

iisakkirotko (changes requested):
Only minor comments, otherwise looks good!

solara/server/jupyter/solara.py
maartenbreddels: Yes, that was a leftover comment, I'll remove it. Thanks

PR #814, opened 2024-10-14
feat: enable navigation to hashes

iisakkirotko:
Previously this was not officially supported, and would only work when moving within a page, or when initially loading a page with SSG enabled.

### DIFF
--- a/solara/server/static/main-vuetify.js
+++ b/solara/server/static/main-vuetify.js
@@ -138,15 +138,25 @@ async function solaraInit(mountId, appName) {
     });
 
     window.addEventListener('solara.router', function (event) {
+        app.$data.urlHasChanged = true;
         if(kernel.status == 'busy') {
             app.$data.loadingPage = true;
         }
     });
     kernel.statusChanged.connect(() => {
+        // When navigation is triggered from the front-end, kernel.status becoming busy and
+        // solara.router event happen in a different order than when navigating through Python, so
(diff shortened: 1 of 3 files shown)


solara/server/static/main-vuetify.js
```diff
@@ -127,7 +127,7 @@ async function solaraInit(mountId, appName) {
             window.navigator.sendBeacon(close_url);
         }
     });
-    let kernel = await solara.connectKernel(solara.jupyterRootPath, kernelId)
+    let kernel = await solara.connectKernel(solara.rootPath + '/jupyter', kernelId)
```
maartenbreddels, line 130: This is a revert of a recent change, maybe accidental overwrite?

maartenbreddels:
If I go:
http://localhost:8765/documentation/advanced/enterprise/oauth
and click on say 'Configuring Fief' i navigate to it, but the hash does not change in the address bar (80% of the time).
If I then click it again, the hash does change. 
I scroll up a bit, and click the https link with the hash to 
http://localhost:8765/documentation/getting_started/deploying/self-hosted#https
That works seems to work, as well as navigating back, but it might be good to check after fixes.
I wonder if the problem is related to having 2 state attributes?

iisakkirotko:
### Merge activity

* **Oct 24, 11:04 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/814).

PR #832, opened 2024-10-24
docs: fixes for mobile users

iisakkirotko:
Because we previously used a `solara.v.NavigationDrawer` with `v_model=True`, that was only hidden and not destroyed on mobile, Vuetify would think that a floating navigation drawer is open on page load. This would mean that the page could not be used before the user tapped somewhere on the page to "close the drawer". Also adds another maximum width restriction to make sure content doesn't become wider than the viewport on mobile resolutions.

### DIFF
--- a/solara/server/assets/style.css
+++ b/solara/server/assets/style.css
@@ -124,6 +124,7 @@ div.highlight {
 
 .solara-autorouter-content {
     height: 100%;
+    max-width: 100%;
 }
 
 /* originally from index.css */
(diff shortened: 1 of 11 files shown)


iisakkirotko:
### Merge activity

* **Oct 28, 4:51 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/widgetti/solara/832).

PR #840, opened 2024-10-31
Docs update cross filter dataframe

lizzy985:


maartenbreddels:
@lizzy985 the code quality workflow error gives a hint on what you need to fix.

lizzy985:
> @lizzy985 the code quality workflow error gives a hint on what you need to fix.
Thank you! I will do a new pull request. But I have a confusion. 

I initially remove `solara.column`. When I did rebase before I committed, there is a conflict. I seems that someone has correct the code before me and you approved his pull request (Am I correct?). When I did a merge, I though what you approved should be good, so I choose to accept what you approved. But apparently, it still should remove `solara.column` as you told me.

maartenbreddels:
There is not need to do new pull requests. If you rebase on master, you can indeed get conflicts, which you need to solve. With a bit of practice you get used to it.

The approval of the other PR is indepandent of this one. Make sure that if you make a new branch, that you always start from master, and not the previous branch. I hope that helps.

Issue #782, opened 2024-09-11
Don't allow boolean operators on reactive variables


maartenbreddels:
Very often, i use
```python
some_var = solara.reactive(0)

if some_var: # OOPS, forgot .value
  ...
```

I don't think there is a reason to do reactive equals comparison, and I think we should error on that. If you want to know if it is the same object, use `is`, and otherwise compare .value.

PR #846, opened 2024-11-01
feat: do not allow reactive to be used as boolean

maartenbreddels:


maartenbreddels:
@Jhsmit i guess you would like this

Jhsmit:
see also #782

solara/toestand.py
```diff
@@ -97,6 +97,35 @@ def __init__(self, merge: Callable = merge_state, equals=equals_extra):
         self.equals = equals
         self.listeners: Dict[str, Set[Tuple[Callable[[T], None], Optional[ContextManager]]]] = defaultdict(set)
         self.listeners2: Dict[str, Set[Tuple[Callable[[T, T], None], Optional[ContextManager]]]] = defaultdict(set)
+        self._prevent_boolean = solara.settings.main.prevent_reactive_boolean
+
+    # make sure all boolean operations give type errors
+    def __bool__(self):
+        if self._prevent_boolean:
+            raise TypeError("Reactive vars are not allowed in boolean expressions, did you mean to use .value?")
+        return True
```
maartenbreddels, line 106:
```suggestion
    if solara.settings.main.prevent_reactive_boolean:
      def __bool__(self):
          if self._prevent_boolean:
              raise TypeError("Reactive vars are not allowed in boolean expressions, did you mean to use .value?")
          return True
```

This may look funny at first, but that does not require up to give a default implementation. The other solution would be to call super(self).__bool__()

maartenbreddels:
Should we start adding a run of the unittests with this setting on? Can be a 2nd commit on this branch.

maartenbreddels:
Note: this feature is not enabled by default in solara 1.x, it will be planned to be the default in 2.x and above.

PR #855, opened 2024-11-08
Add timeout to pypi version request, preventing undue hangups on startup

### LINKED ISSUE #853: Add a short timeout to `_check_version()` requests call
Some corporate proxies deny access to `pypi.org`. In these cases, `solara run app.py` hangs while it tries to make contact with pypi to check whether its version is up-to-date. Ideally, the call to `requests.get()` occurs with `timeout=1` or something very short & reasonable to prevent inordinate app load times, without requiring an explicit `--no-check-version` (which is unintuitive without debugging)

ntjess:
Fixes #853

### DIFF
--- a/solara/__main__.py
+++ b/solara/__main__.py
@@ -75,7 +75,9 @@ def _check_version():
     import requests
 
     try:
-        response = requests.get("https://pypi.org/pypi/solara/json")
+        # Since we already catch the bare-except below, no need to explicitly capture
+        # requests.exceptions.Timeout here
+        response = requests.get("https://pypi.org/pypi/solara/json", timeout=0.2)
         latest_version = response.json()["info"]["version"]
     except:  # noqa: E722
         return
--- a/solara/__main__.py
+++ b/solara/__main__.py
@@ -21,6 +21,8 @@ import solara.server.threaded
 
 from .server import telemetry
 
+print_mutex = threading.Lock()
+
 try:
     from solara_enterprise.ssg import ssg_crawl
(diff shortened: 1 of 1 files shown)


maartenbreddels:
We thought a bit about it. What do you think of this solution? Where we only show it if we found a new version on time.

ntjess:
Have you tested the timing between the browser opening and the message being printed in the console?

My only concern with this new approach is the following scenario:
- User runs `solara run ...`
- Browser window opens, user no longer thinks about their terminal
- Upgrade message prints, but user never sees it because their focus is already on the browser
- App generates `print()` / logging statements that obfuscate the upgrade message
- User never learns there is a solara update

ntjess:
Another thought: We may also want to read the pip conf (through an approach like `pip config list` -> check if `index-url` is overwritten) to see which URL should be hit instead of defaulting to pypi.

This is a more robust solution than just a timeout, and would work for e.g. private pypi setups like artifactory. However, it is nontrivial to know ahead of time where pip will fetch from (overloads at global/user level, even per-pip command). So maybe the complexity outweighs the benefits.

maartenbreddels:
I revered to your original idea, but we now execute it in a thread, so it is never delaying on the terminal. We still add a timeout, so the thread doesn't hang around (500ms should be enough, overwise indeed, focus is probably shifted).

Using a Lock/mutex, we make sure the printing of text does not mix.

maartenbreddels:
Thanks @ntjess !

PR #876, opened 2024-11-25
Added a httponly option for HTMLResponse.set_cookie

lopezvoliver:
Following the discussion in #799, I added an option configurable through an environment variable `SOLARA_SESSION_HTTP_ONLY`, and passing this to the `httponly` option in [HTMLResponse](https://www.starlette.io/responses/). 

As discussed, this change is backwards compatible because it defaults to False.

Here we can verify the httpOnly attribute of the `solara-session-id` cookie using Developer tools:

image

and here's another screenshot showing the behavior when setting the environment variable `SOLARA_SESSION_HTTP_ONLY=True`:

image

Note that the cookie no longer appears in `document.cookie`. 

Thus, this PR makes it possible for a solara-based application to comply with [cwe 402](https://cwe.mitre.org/data/definitions/402.html)

### DIFF
--- a/solara/server/settings.py
+++ b/solara/server/settings.py
@@ -132,6 +132,7 @@ OAUTH_TEST_CLIENT_IDs = [AUTH0_TEST_CLIENT_ID, FIEF_TEST_CLIENT_ID]
 
 class Session(BaseSettings):
     secret_key: str = SESSION_SECRET_KEY_DEFAULT
+    http_only: Optional[bool] = None
     https_only: Optional[bool] = None
     same_site: str = "lax"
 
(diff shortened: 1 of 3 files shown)


iisakkirotko (approved):
Looks good to me. Thanks a lot for the PR @lopezvoliver!

solara/server/settings.py
```diff
@@ -132,6 +132,7 @@ class Config:
 
 class Session(BaseSettings):
     secret_key: str = SESSION_SECRET_KEY_DEFAULT
+    http_only: Optional[bool] = None
```
maartenbreddels, line 135:
```suggestion
    http_only: bool = False
```

I think we can do it like this. The reason https_only is 'tri-state' is that we do some checks when it is not explicitly set.

maartenbreddels (changes requested):
Awesome, I think we can make it a bit simpler even, what do you think?

lopezvoliver:
> Awesome, I think we can make it a bit simpler even, what do you think?

Yes, and this is also consistent with the default False value in [starlette](https://github.com/encode/starlette/blob/13d0c1fbcdb6c6f861155beed352112843e57717/starlette/responses.py#L97).

maartenbreddels:
Thanks for your contribution @lopezvoliver !

PR #878, opened 2024-11-29
docs: use mkdocs-pycafe to include links to pycafe

maartenbreddels:
This renders like:

image

### DIFF
--- a/solara/components/markdown.py
+++ b/solara/components/markdown.py
@@ -1,10 +1,12 @@
+import functools
 import hashlib
 import html
 import logging
 import textwrap
 import traceback
 import warnings
-from typing import Any, Dict, List, Union, cast
+from typing import Any, Dict, List, Optional, Union, cast
+import typing
 
 import ipyvuetify as v
 
@@ -29,6 +31,9 @@ except ModuleNotFoundError:
 else:
     from pygments.formatters import HtmlFormatter
     from pygments.lexers import get_lexer_by_name
+
+if typing.TYPE_CHECKING:
+    import markdown
 
 logger = logging.getLogger(__name__)
 
@@ -231,7 +236,7 @@ module.exports = {
     return template
 
 
-def _highlight(src, language, unsafe_solara_execute, extra, *args, **kwargs):
+def _highlight(src, language, unsafe_solara_execute, *args, **kwargs):
     """Highlight a block of code"""
 
     if not has_pygments:
@@ -256,6 +261,10 @@ def _highlight(src, language, unsafe_solara_execute, extra, *args, **kwargs):
             return src_html + html_no_execute_enabled
     else:
         return src_html
+
+
+def formatter(unsafe_solara_execute: bool):
+    return functools.partial(_highlight, unsafe_solara_execute=unsafe_solara_execute)
 
 
 @solara.component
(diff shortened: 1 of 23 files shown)


solara/components/markdown.py
```diff
@@ -231,7 +236,7 @@ def _markdown_template(
     return template
 
 
-def _highlight(src, language, unsafe_solara_execute, extra, *args, **kwargs):
+def _highlight(src, language, unsafe_solara_execute, extra=None, *args, **kwargs):
```
iisakkirotko, line 239: Is this change related or needed? It doesn't seem like we use `extra` anywhere

solara/components/markdown.py
```diff
@@ -333,6 +342,7 @@ def Page():
      * `unsafe_solara_execute`: If True, code marked with language "solara" will be executed. This is potentially unsafe
         if the markdown text can come from user input and should only be used for trusted markdown.
      * `style`: A string or dict of css styles to apply to the rendered markdown.
+     * `md`: A markdown object to use for rendering. If not provided, a markdown object will be created.
```
iisakkirotko, line 345: Clarify how case with no `md` works

```suggestion
     * `md`: A markdown object to use for rendering. If not provided, a markdown object will be created based on provided `md_text`.
```

iisakkirotko (changes requested):
If I try to visit /documentation/getting_started, I get the following error:
```
Traceback (most recent call last):
  File "/Users/iisakki/micromamba/envs/solara-dev/lib/python3.10/site-packages/reacton/core.py", line 1707, in _render
    root_element = el.component.f(*el.args, **el.kwargs)
  File "/Users/iisakki/Code/widgetti/solara/solara/website/components/markdown.py", line 72, in MarkdownWithMetadata
    md = solara.use_memo(make_markdown_object, dependencies=[unsafe_solara_execute])
  File "/Users/iisakki/micromamba/envs/solara-dev/lib/python3.10/site-packages/reacton/core.py", line 879, in use_memo
    return rc.use_memo(f, dependencies, debug_name)
  File "/Users/iisakki/micromamba/envs/solara-dev/lib/python3.10/site-packages/reacton/core.py", line 1244, in use_memo
    value = f()
  File "/Users/iisakki/Code/widgetti/solara/solara/website/components/markdown.py", line 59, in make_markdown_object
    "format": mkdocs_pycafe.formatter(type="solara", next_formatter=formatter(unsafe_solara_execute), inside_last_div=False),
TypeError: formatter() got an unexpected keyword argument 'next_formatter'
```
Also, since dependencies have changed, I think the CI package locks should be deleted to be regenerated.

solara/components/markdown.py
maartenbreddels: No, the current text is correct, we'll create the markdown object, which is then used to render the text provided by md_text. So it does not replace md_text.
Could we express this more clearly? Or do you think this is ok?

iisakkirotko: I see. In that case maybe we could change the argument name and description to refer to it as a parser, i.e. `md_parser`? This is how the `Markdown` class is specified in https://github.com/Python-Markdown/markdown/blob/0b5e80efbb83f119e0e38801bf5b5b5864c67cd0/markdown/core.py#L48

Although in their docs they only refer to `markdown.Markdown` instances.

maartenbreddels:
This uses the pycafe mkdocs plugin:
https://mkdocs.py.cafe/

In the documentation, we use it like this:
````
```{.python pycafe-link extra-requirements="vega_datasets"}
...
```
````

Nowadays we can also use:
````
```python {pycafe-link extra-requirements="vega_datasets"}
...
```
````


For the cases where we render the live output, we should use something like:

````
```solara {pycafe-link extra-requirements="vega_datasets"}
...
```
````

If we don't need extra requirements, it should look like

````
```solara {pycafe-link}
...
```
````

solara/components/markdown.py
maartenbreddels: Yes, I like `md_parser`!

PR #891, opened 2024-12-02
docs update code vbox

lizzy985:


iisakkirotko (changes requested):
`solara.Vbox` has also been deprecated. Could you add a deprecation notice to this page? See for example the component definition https://github.com/widgetti/solara/blob/a980aa5ee602f00bae818ff8debbb9b73ca58dd0/solara/components/misc.py#L139

PR #893, opened 2024-12-02
docs update code markdown_editor

lizzy985:


solara/website/pages/documentation/components/output/markdown_editor.py
```diff
@@ -39,9 +38,9 @@ def Page():
     with solara.HBox(grow=True) as main:
         with solara.VBox():
```
iisakkirotko, line 39: I think these should also be replaced with `Row` and `Column`.

iisakkirotko (changes requested):

iisakkirotko (approved):

solara/website/pages/documentation/components/output/markdown_editor.py
```diff
@@ -36,12 +35,12 @@ def Page():
     """.strip()
 
     markdown_text, set_markdown_text = solara.use_state(markdown_initial)
-    with solara.HBox(grow=True) as main:
-        with solara.VBox():
+    with solara.Row(grow=True) as main:
```
iisakkirotko, line 38: `Row` doesn't accept `grow` as an argument.

```suggestion
    with solara.Row() as main:
```

iisakkirotko, line 38: Also, I missed this last time, but the `return main` pattern can also be removed here.

```suggestion
    with solara.Row(grow=True):
```

solara/website/pages/documentation/components/output/markdown_editor.py
```diff
@@ -36,12 +35,12 @@ def Page():
     """.strip()
 
     markdown_text, set_markdown_text = solara.use_state(markdown_initial)
```
iisakkirotko, line 37: And I also now noticed that the `use_state` here can be replaced with `use_reactive`, and `.value` and `.set` can be used throughout.

iisakkirotko (changes requested):

PR #894, opened 2024-12-02
docs update code markdown

lizzy985:


solara/website/pages/documentation/components/output/markdown.py
```diff
@@ -43,12 +42,12 @@ def Page():
         with solara.VBox():
```
iisakkirotko, line 42: Replace this with `Column` as well.

solara/website/pages/documentation/components/output/markdown.py
```diff
@@ -43,12 +42,12 @@ def Page():
...
             solara.Markdown("# Input text")
             with solara.Padding(2):
-                with rv.Sheet(elevation=2):
-                    rv.Textarea(v_model=markdown_text, on_v_model=set_markdown_text, rows=30)
+                with solara.v.Sheet(elevation=2):
+                    solara.v.Textarea(v_model=markdown_text, on_v_model=set_markdown_text, rows=30)
```
iisakkirotko, line 46: We also now have `solara.InputTextArea`, I think it makes sense to replace `solara.v.Textarea` with that.

iisakkirotko (changes requested):

solara/website/pages/documentation/components/output/markdown.py
iisakkirotko: Looks like this wasn't addressed yet.

Issue #841, opened 2024-10-31
Webhooks - example


JovanVeljanoski:
Hi, 

Would it be possible to add the simplest possible example of how one would create or "listen to" webhooks with solara? 
Thank you!

JovanVeljanoski:
In addition, would be nice if we can add additional (custom) endpoints to the solara server, and implement it into the app.

JovanVeljanoski:
Ok, i am pretty sure I am doing something very unholy here, but let me share an example.
(based on https://github.com/widgetti/solara/pull/670)

```python

from copy import deepcopy

import solara
import solara.lab
from starlette.responses import JSONResponse
import asyncio



added = False

webhooks_result = solara.reactive('initial value')


async def update_webhook_value(message):
    old_value = webhooks_result.value
    message = deepcopy(message)
    print(f'Updating webhook value from "{old_value}" to "{message}"')
    webhooks_result.set(message)
    print(f'Updated webhook value to: "{webhooks_result.value}"')


async def webhook_handler(request):
    try:
        payload = await request.json()
        print('Request received!')
        print(payload.get('message', 'no message'))
        payload = deepcopy(payload)
        asyncio.create_task(update_webhook_value(payload.get('message', 'no message')))
        return JSONResponse({
            "status": "success",
            "message": "Webhook received",
            "data": payload
        })
    except Exception as e:
        return JSONResponse({
            "status": "error",
            "message": str(e)
        }, status_code=400)


def add_webhook_handler():
    global added
    if not added:
        import solara.server.starlette
        solara.server.starlette.app.router.add_route("/webhook", webhook_handler, methods=["POST"])
        added = True


@solara.component
def Page():
    add_webhook_handler()
    value = solara.use_reactive('')

    def fetch_webhook_value():
        print('Fetching webhook value')
        print(f'The webhook value is: {webhooks_result.value}')
        value.set(webhooks_result.value)
        webhooks_result.set('what is this?')

    with solara.Column(align="center"):
        with solara.Card(style="width: 400px; margin: auto;"):
            solara.Text("Hello")
            solara.Markdown("This is a markdown block")
            solara.Markdown(f"Webhook added: {added}")


            solara.Markdown("### Latest Webhook Data")
            solara.Markdown(f'{webhooks_result.value}')
            solara.Button(label='Fetch webhook value', on_click=fetch_webhook_value)
            solara.Text(f'Value: {value.value}')
```

Then one can hit the endpoint with 

```curl
[...]

PR #919, opened 2024-12-12
fix: a failing websocket.send would drop the connection silently

maartenbreddels:
Instead of assuming all exceptions are due to a closed connection, we only ignored the exception when it is due to a closed connection.

This caused issues in #841 where we called from the same thread as the websocket's portal, which caused the connection to be ignored without any error message.

### DIFF
--- a/solara/server/starlette.py
+++ b/solara/server/starlette.py
@@ -15,6 +15,7 @@ import anyio
 import starlette.websockets
 import uvicorn.server
 import websockets.legacy.http
+import websockets.exceptions
 
 from solara.server.utils import path_is_child_of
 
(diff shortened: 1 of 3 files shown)


solara/server/starlette.py
```diff
@@ -121,9 +122,31 @@ async def process_messages_task(self):
...
+    async def _send_bytes_exc(self, data: bytes):
+        # make sures we catch the starlette/websockets specific exception
+        # and re-raise it as a websocket.WebSocketDisconnect
+        try:
+            await self.ws.send_bytes(data)
+        except websockets.exceptions.ConnectionClosed as e:
+            raise websocket.WebSocketDisconnect() from e
+        except RuntimeError as e:
+            # starlette throws a RuntimeError once you call send after the connection is closed
+            raise websocket.WebSocketDisconnect() from e
```
iisakkirotko, line 138: Should this be implemented as a decorator / context manager to not define two almost identical functions?

iisakkirotko (approved):
Looks good. Only one comment about the implementation of the wrapper methods, but if it let's us get this in faster, we can merge as is 👍

solara/server/starlette.py
maartenbreddels: As discussed in private, it's DRY (don't repeat yourself) vs YAGNI (you ain't gonna need it). It would  make it more difficult to read, vs it makes it more difficult to maintain.
In this particular case, we probably should have overwritten the `send(..)` method and should have pushed the branching down.

PR #924, opened 2024-12-13
refactor: change the order of importing/executing solara script

maartenbreddels:
We now do this after solara.server.starlette is imported, making the starlette app variable accessible from the solara script or module. This makes it easier to add custom endpoints to a solara app.

This will make the script in #841 much simpler, as well as #670

### DIFF
--- a/solara/server/app.py
+++ b/solara/server/app.py
@@ -62,6 +62,35 @@ class AppScript:
         else:
             self.name = name
         self.path: Path = Path(self.name).resolve()
+        if self.path.is_dir():
+            self.type = AppType.DIRECTORY
+            # resolve the directory, because Path("file").parent.parent == "." != ".."
+            self.directory = self.path.resolve()
+        elif self.name.endswith(".py"):
+            self.type = AppType.SCRIPT
+            self.directory = self.path.parent.resolve()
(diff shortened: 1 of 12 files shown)


solara/server/app.py
```diff
@@ -110,45 +138,26 @@ def add_path():
...
             reload.reloader.watcher.add_file(self.path)
-            self.directory = self.path.parent.resolve()
             initial_namespace = {
                 "__name__": "__main__",
             }
             with reload.reloader.watch():
                 routes = [solara.autorouting._generate_route_path(self.path, first=True, initial_namespace=initial_namespace)]
-        elif self.name.endswith(".ipynb"):
-            self.type = AppType.NOTEBOOK
+        elif self.type == AppType.NOTEBOOK:
```
iisakkirotko, line 150: This case and `self.type == AppType.SCRIPT` are almost identical, maybe their cases here can be merged?

iisakkirotko (approved):
Looks good to me! Feel free to merge regardless of my comment (again, we can always do that refactor later)

solara/server/app.py
maartenbreddels: Good catch, but I'd rather keep it like this, I don't think the code will look cleaner in the end

PR #930, opened 2024-12-15
fix: close kernels on shutdown of asgi/starlette server

maartenbreddels:
Since Flask does not have lifecycle management, we cannot do the same for Flask.

We could implement this for Flask by listening to signals, but that's not the responsibility of the solara server.

Fixes #929

### DIFF
--- a/solara/server/starlette.py
+++ b/solara/server/starlette.py
@@ -576,6 +576,12 @@ def on_startup():
 
 
 def on_shutdown():
+    # shutdown all kernels
+    for context in list(kernel_context.contexts.values()):
+        try:
+            context.close()
+        except:  # noqa
+            logger.exception("error closing kernel on shutdown")
     telemetry.server_stop()
 
 


solara/server/starlette.py
```diff
@@ -576,6 +576,12 @@ def on_startup():
 
 
 def on_shutdown():
+    # shutdown all kernels
+    for context in list(kernel_context.contexts.values()):
+        try:
+            context.close()
+        except:  # noqa
+            logger.exception("error closing kernel on shutdown")
```
iisakkirotko, line 584: Would it be useful to identify which kernel's shutdown failed for logging purposes?

iisakkirotko (approved):
Feel free to merge. Just the one question (and we can always implement that later)

solara/server/starlette.py
maartenbreddels: I don't think the ID gives much information, but this will show the stacktrace, which should give a hint.

Issue #932, opened 2024-12-16
internal `_use_input_type` hook keeps old functions


iisakkirotko:
## Expected Behavior

If I change a variable that is used in one of the functions given to `_use_input_type`, I expect the output from the hook to change. However, as of current this isn't supported

## Current Behavior

If I change one of the variables, the old value is still used within the hook.

## Steps to Reproduce the Problem

Go to 
https://py.cafe/iisakkirotko/solara-date-format-changer
And click the button

## Fix

We should add a dependency array as an argument to `_use_input_type`, so that we could remove and re-run the effect with the new values when one of the used arguments changes

iisakkirotko:
Noticed this issue while working on https://github.com/widgetti/solara/pull/654, although it does also apply to `InputDate`

PR #933, opened 2024-12-16
fix: inputdate not allowing format changes after creation

iisakkirotko:
### All Submissions:

* [x] I installed `pre-commit` prior to committing my changes (see [development setup docs](https://solara.dev/documentation/advanced/development/setup#contributing)).
* [x] My commit messages conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
* [x] My PR title conforms to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
* [x] I linked to any relevant issues.

### Description of changes

Previously, as seen in https://py.cafe/iisakkirotko/solara-date-format-changer, `InputDate` would not respect `date_format` changes after the component was created. This would mean that if the format was changed, any input through the picker would be labeled invalid. Also fixes the same issue for `InputTime` (NOTE: Requires https://github.com/widgetti/solara/pull/654 to be merged in order to merge correctly. Graphite didn't like the fact that #654 comes from another repo)

Fixes https://github.com/widgetti/solara/issues/932

### DIFF
--- a/solara/components/input.py
+++ b/solara/components/input.py
@@ -357,6 +357,7 @@ def _use_input_type(
     parse: Callable[[Optional[str]], T],
     stringify: Callable[[Optional[T]], str],
     on_value: Union[None, Callable[[Optional[T]], None], Callable[[T], None]] = None,
+    extra_dependencies=[],
 ):
     reactive_value = solara.use_reactive(input_value, on_value)  # type: ignore
     del input_value, on_value
@@ -373,6 +374,10 @@ def _use_input_type(
         error_message = str(e.args[0])
 
     def sync_back_input_value():
+        # Make sure we update string_value when the effect is rerun,
+        # Since the parsing & stringigying functions might have changed
+        set_string_value(stringify(reactive_value.value) if reactive_value.value is not None else None)
+
         def on_external_value_change(new_value: Optional[T]):
             new_string_value = stringify(new_value)
             try:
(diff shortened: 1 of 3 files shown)


solara/components/input.py
```diff
@@ -386,7 +391,7 @@ def on_external_value_change(new_value: Optional[T]):
 
         return reactive_value.subscribe(on_external_value_change)
 
-    solara.use_effect(sync_back_input_value, [reactive_value])
+    solara.use_effect(sync_back_input_value, [reactive_value, *extra_dependencies])
```
maartenbreddels, line 394: This should be `reactive_value.value` and maybe instead of extra_dependencies, use stringify.

I think our equals is good enough to compare separate function which have the same closure and code to be the same.

iisakkirotko: I think not using `reactive_value.value` is correct here since the effect subscribes to changes on `reactive_value`. Since the `reactive_value` is declared at hook level I don't think having it in the dependencies ever causes a rerun of the effect, so we could remove it for clarity?

maartenbreddels: Ah true!

iisakkirotko:
* **#933** <a href="https://app.graphite.dev/github/pr/widgetti/solara/933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/widgetti/solara/933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

PR #935, opened 2024-12-17
fix: value of a reactive value can be None (use placeholder)

maartenbreddels:


tests/unit/task_test.py
```diff
@@ -387,32 +388,32 @@ def square_minus_one():
...
         assert square_minus_one.value == 35
         assert called == 2
-        assert len(square._result._storage.listeners2["t2"]) == 1
+        assert len(get_storage(square._result).listeners2["t2"]) == 1
         # square_minus_one._auto_subscriber.value.reactive_used == {square.value}
 
     with context1:
         assert r1 is square._result
-        # assert len(square.result._storage.listeners2["t1"]) == 1
+        # assert len(square.result.listeners2["t1"]) == 1
```
iisakkirotko, line 416: I'm guessing this can be either removed, or adapted to use `get_storage(...)`?

```suggestion
```

iisakkirotko (commented):
Otherwise looks good! Once (🤞) the tests pass this should be good to go.

tests/unit/task_test.py
maartenbreddels: I was doing a replace-all, and not sure what I intend here, so i left it here fore future us to take a look at.

PR #940, opened 2024-12-20
feat: transitions component for slick UI's

maartenbreddels:
Based on Vue's transitions

TODO
 * [ ] add example to pycafe + CI curl test (draft is already at https://py.cafe/solara/transitions )
 * [ ] add howto add transitions


rileythai:
Do we plan to support structures like this for transitions?
```python
for (i,data) in dataclass.property:
    Component(data=data).key(i) # iteratively generated
```
where `Component` is some Solara component that is generated from the dataclass.

Vuetify v2 docs has something similar [here](https://v2.vuetifyjs.com/en/styles/transitions/#todo-list).  My initial try to implement a parallel format with `transition-group` failed but I think I had it wrong.

PR #941, opened 2024-12-20
Feat chatinput improvements

maartenbreddels:
two small improvements.
 * autofocus
 * disable send button, not the input

This will make the CI pycafe run fail, but we can only fix that after a release.


iisakkirotko (approved):
Looks good to me!

Issue #927, opened 2024-12-13
get_widget can fail


maartenbreddels:
Although a reacton bug, it's happening in a solara context, so I thought for visibility it is better to post here.

[...]

PR #997, opened 2025-02-07
feat: Bokeh support as FigureBokeh

### LINKED ISSUE #202: Feature Request: Bokeh support
Just wanted to request Bokeh integration. I realize it can be somewhat supported with the `display` functionality, but having integration would be nice to get analogous functionality to Altair and Plotly.

https://bokeh.org/

rileythai:
### All Submissions:
* [X] docs/examples
* [X] pre-commit
* [ ] tests
### TODO
* [X] fix printing issue
* [X] document common issues and their relevant workarounds
### Description of changes

This PR resolves #202, and adds a `solara.lab` component to support Bokeh `figure` and `Plot` objects. It additionally adds dependencies for `bokeh` and `jupyter_bokeh`. 

Part of the code is based on the implementation written in #202 by @maartenbreddels.

It behaves similarly to `FigurePlotly`. It takes arguments of `light_theme` and `dark_theme` to update the theme via a solara-side `effect` (the document instance must exist for us to update theming via the high-level interface)

This PR adds:
```
├── solara
│  └── lab
│     └── components
│        ├── bokehloaded.vue
│        └── figurebokeh.py
└───── website
      ├── pages
      │  ├── apps
      │  └────  scatter-bokeh.py
      └──── documentation
            └── examples
               ├── fullscreen
               │  └── scatter_bokeh.py
               └── visualization
                  └── bokeh.py
```

and modifies relevant files for importing components in `lab`

### Known issues/pitfalls
- Bokeh does not support an axis type change, so changing axes via high-level interface `x_axis_type`, etc doesnt work.
  - log workaround: update `x_scale` and/or `y_scale` of figure via a `use_effect`
  - categorical workaround: scale/map data server-side + update tickers directly, see the scatter-bokeh.py example
 - JS-side selection events on a `Tool` will fail to link to attributes of any `bokeh.model` if it is regenerated with each render.
   - occurs `bokeh.model.tools` instance with a callback like `CustomJS(args=dict(object=object))`, such as `object.selected.indices` or similar from a `ColumnDataSource` upon a selection.
    - suggested approach: `use_memo` models with callbacks to exclude from regeneration, then use `use_effect` or similar for attribute updates.


solara/website/pages/documentation/examples/visualization/bokeh.py
```diff
@@ -0,0 +1,48 @@
...
+title = "Scatter plot using Bokeh"
+
+df = penguins.data
+
+
+@solara.component
+def Page():
+    all_species = df["species"].unique().tolist()
+    species = solara.use_reactive(all_species[0])
+    with solara.Div() as main:
```
maartenbreddels, line 24:
```suggestion
    with solara.Div():
```

solara/website/pages/documentation/examples/visualization/bokeh.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+        # add a scatter, colorbar, and mapper
+        mapper = LinearColorMapper(palette="Viridis256", low=dff["body_mass_g"].min(), high=dff["body_mass_g"].max())
+        cb = ColorBar(color_mapper=mapper, title="Body mass [g]")
+        p.scatter(source=source, x="x", y="y", marker="circle", size=8, fill_color={"field": "z", "transform": mapper})
+        p.add_layout(cb, "right")
+
+        solara.lab.FigureBokeh(p, dark_theme="carbon", dependencies=[species])
+    return main
```
maartenbreddels, line 48:
```suggestion
```

No need to use main/return main

solara/lab/components/figurebokeh.py
```diff
@@ -0,0 +1,73 @@
...
+    pass
+
+
+def FigureBokeh(
+    fig,
+    dependencies=None,
+    light_theme: str | Theme = "light_minimal",
+    dark_theme: str | Theme = "dark_minimal",
+):
+    # NOTE: no docstring because not a component.
```
maartenbreddels, line 23: Why is this not a component? Because if we use this function conditionally, we'll have conditional hooks

maartenbreddels (commented):
Nice! I think @Jhsmit will also like this. Maybe @Jhsmit can also review this?

solara/lab/components/figurebokeh.py
rileythai: It's the same reason `FigurePlotly` isn't a component. If you want to attach an effect callback to the figure widget, you won't be able to access it with `sl.get_widget` outside of the function itself.

If you make it a component, you'll get an error on `sl.get_widget` because it searches for the `FigureBokeh` and  not the `BokehModel` instance  
```python
 raise KeyError(f"Element {el} not found in all known widgets")  # for the component {context.widgets}")
KeyError: "Element FigureBokeh(Plot(id='p2331', ...), ...) not found in all known widgets"
```

rileythai: I end up running into issues with https://github.com/widgetti/solara/issues/927 when using an effect outside if it is a component.

rileythai: I ran into issues relating to #927 on occassion, couldn't track down what caused it

PR #1003, opened 2025-02-11
Feature Request: Add custom labels to the AppLayout Tab

### LINKED ISSUE #1002: Feature Request: Add custom labels to the AppLayout Tab
## Feature Request

- [ ] The requested feature would break current behaviour
- [x] I would be interested in opening a PR for this feature

### What problem would this feature solve? Please describe.
Add a custom label to the AppLayout Tab of the default route "/" or other routes if "label" is defined. Currently, the label of the default route is always "Home".
```
routes = [
    solara.Route("/", component=Page, label="Foo", layout=Layout),
    solara.Route("test", component=Test, label="Test123"),
    solara.Route("example", component=Example)
]
```

### Describe the solution you'd like
If the user defines the "label" parameter of a route, the AppLayout Tab should use this naming. In the above examples the AppLayout tabs labels would be "Foo", "Test123" and "example".

johannesWen:
closes #1002

### DIFF
--- a/solara/components/applayout.py
+++ b/solara/components/applayout.py
@@ -282,7 +282,10 @@ def AppLayout(
     if (tabs_element is None) and routes and navigation and (len(routes) > 1):
         with solara.lab.Tabs(value=index, on_value=set_path, align="center") as tabs_element:
             for route in routes:
-                name = route.path if route.path != "/" else "Home"
+                if route.label:
+                    name = route.label
+                else:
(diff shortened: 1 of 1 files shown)


iisakkirotko:
Hey! Sorry for the slow reply, we've all been pretty busy. This is already included in https://github.com/widgetti/solara/pull/657, although I have to say I prefer your syntax. Let's hope we have some more time to address the rest of the 2.0 PRs soon.

PR #1036, opened 2025-04-25
fix: follow asgi spec and also check if text or bytes are None (Modal fix)

maartenbreddels:
We assumed when the text or bytes were None that it was a missing value, but on Modal we do see a message like that, e.g.:
{'type': 'websocket.receive', 'bytes': b'...', 'text': None}}

See https://asgi.readthedocs.io/en/latest/specs/www.html#receive-receive-event which explain that None and missing are equivalent.

### DIFF
--- a/solara/server/starlette.py
+++ b/solara/server/starlette.py
@@ -230,9 +230,9 @@ await to_thread.run_sync(my_update)
                 fut = self.portal.spawn_task(self.ws.receive)
 
             message = await asyncio.wrap_future(fut)
-        if "text" in message:
+        if message.get("text") is not None:
             return message["text"]
-        elif "bytes" in message:
+        elif message.get("bytes") is not None:
             return message["bytes"]
         elif message.get("type") == "websocket.disconnect":
             raise websocket.WebSocketDisconnect()


solara/server/starlette.py
```diff
@@ -230,9 +230,9 @@ async def receive(self):
                 fut = self.portal.spawn_task(self.ws.receive)
 
             message = await asyncio.wrap_future(fut)
-        if "text" in message:
+        if "text" in message and message["text"] is not None:
```
Ben-Epstein, line 233: Fwiw
`if message.get("text") is not None`

solara/server/starlette.py
```diff
@@ -230,9 +230,9 @@ async def receive(self):
...
             return message["text"]
-        elif "bytes" in message:
+        elif "bytes" in message and message["bytes"] is not None:
```
Ben-Epstein, line 235: Same here

Ben-Epstein (approved):

solara/server/starlette.py
maartenbreddels: yeah, I agree

maartenbreddels:
Failure is:
```
FAILED tests/integration/ssg_test.py::test_ssg[flask-chromium] - playwright._impl._errors.Error: Page.goto: Navigation to "http://localhost:18765/" is interrupted by another navigation to "http://localhost:18765/"
```
Seeing this a lot on windows only, we should disable this test on windows.

Issue #1053, opened 2025-05-19
Improve display in PyPI Search Display


jacklinke:
When searching PyPI, the `solara` package displays no description, so potential users have no idea what is *is*.

![Image](https://github.com/user-attachments/assets/bc9ee56a-1874-4fee-9014-36bffa8d233c)

Proposing a minor update to `pyproject.toml` to populate the description field on PyPI.

PR #1054, opened 2025-05-19
docs: add description to pyproject.toml

jacklinke:
### All Submissions:

* [x] My commit messages conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
* [x] My PR title conforms to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
* [x] I linked to any relevant issues.

### Changes to / New Features:

* [x] I included the `description` field in `pyproject.toml` per [the Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) completing #1053

### Description of changes

Adding this field will populate the description on PyPI's search page.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "solara-ui"
+description = "A Pure Python, React-style Framework for Scaling Your Jupyter and Web Apps"
 readme = "README.md"
 authors = [{name = "Maarten A. Breddels", email = "maartenbreddels@gmail.com"}]
 license = {file = "LICENSE"}


iisakkirotko (changes requested):
Hey @jacklinke! First off, thanks for the PR. I think two things should be addressed before we can merge this:

1. As you can see the build fails because `description` is listed under `dynamic` in the pyproject.toml. I don't think it needs to be there, since evidently the value doesn't get added dynamically anywhere (If I'm wrong @maartenbreddels feel free to correct me).
2. The pyproject.toml in the project root is actually for the package `solara-ui` (due to how the project was evolving when we split off the `solara` package to `solara-ui`, `solara-server`, etc. So I think this description should go to the file found under `packages/solara-meta`. I think it would also be good to change this in the other packages in this repo (I think all of them have the same issue), but I understand if you feel that's out of your scope. Let me know!

Issue #1110, opened 2025-10-14
Pin `ipykernel` to `<7.0.0` [breaking changes]


chrisji:
A major release of `ipykernel` has been released. We have tracked this as the cause of breaking changes to `solara`.

See: https://github.com/ipython/ipykernel/releases/tag/v7.0.0

Resulting error trace for context:

```
│   File "/usr/local/lib/python3.10/site-packages/comm/base_comm.py", line 130, in close                                                                                                 │
│     self.publish_msg(                                                                                                                                                                  │
│   File "/usr/local/lib/python3.10/site-packages/solara/server/kernel.py", line 102, in publish_msg                                                                                     │
│     parent=self.kernel.get_parent("shell"),                                                                                                                                            │
│   File "/usr/local/lib/python3.10/site-packages/ipykernel/kernelbase.py", line 797, in get_parent                                                                                      │
│     return self._shell_parent.get()                                                                                                                                                    │
│ LookupError: <ContextVar name='shell_parent' at 0x7fec141ce1b0>
```

JovanVeljanoski:
Hi!  I spent some time preparing this but @chrisji beat me to it. In case it is helpful, i prepared some examples and such to illustrate the problem. I  am not sure if it is a breaking change in solara, or we have been abusing it.. 

The ipykernel 7.0.0 breaking change seems to be related to improved async context isolation, which is generally a good thing, but it breaks this previously-working pattern in Solara.

## Questions for Maintainers

1. Is this pattern (calling a Solara task from within another async task after awaits) intended to be supported or was I abusing this and it got lucky because it worked before?
2. Could Solara's task execution mechanism be updated to properly propagate or handle context variables in ipykernel 7.0.0?
3. Are there any best practices or patterns you'd recommend for structuring async tasks that need to update state after async operations?


## Summary

When calling a Solara task from within another async Solara task after `await` operations, the application crashes with a `LookupError` for the `shell_parent` context variable in ipykernel 7.0.0. This pattern works fine in ipykernel 6.x.

## Environment

- **Solara version**: 1.51.1
- **Python version**: 3.12.9
- **ipykernel 6.29**:  Works
- **ipykernel 7.0.0+**:  Fails with `LookupError`

## Minimal Reproducible Example

```python
import solara
import solara.tasks
import asyncio

# Simulated state
app_state_counter = solara.reactive(0)

@solara.tasks.task
def update_progress():
    """A simple sync task that updates state."""
[...]

PR #1111, opened 2025-10-16
Add pre-release dependency testing to catch breaking changes early

EwoutH:
This adds a new CI job that runs nightly to test against pre-release versions of critical dependencies (ipykernel, ipywidgets, ipyvue, ipyvuetify, jupyter-client, traitlets).

The job will create warnings when pre-release versions introduce breaking changes, giving us advance notice before stable releases.

This would have caught the ipykernel 7.0.0 breaking changes that are currently causing test failures (see widgetti/solara#1110).

Only runs on scheduled builds to minimize CI overhead.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -583,6 +583,56 @@ jobs:
           path: ./**/${{ env.LOCK_FILE_LOCATION }}
           include-hidden-files: true
 
+  prerelease-test:
+      needs: [build]
+      timeout-minutes: 25
+      runs-on: ubuntu-latest
+      if: github.event_name == 'schedule'  # Only run on nightly builds
+      strategy:
+        fail-fast: false
+        matrix:
+          python-version: [3.12]
+          ipywidgets_major: ["8"]
+  
+      steps:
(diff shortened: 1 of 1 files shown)


maartenbreddels:
We already do this, although it's not that clearly visible from the workflow file:
https://github.com/widgetti/solara/blob/0f5facd9a513161584792100ed4c4e07a0d036bb/.github/workflows/test.yaml#L301

If we run during the night, we don't use the lockfile, and we indeed saw that the uvicorn dep was breaking (next is the time to fix, which is now :))

maartenbreddels:
ok, this is with --pre actually, hmm, i'm not sure if solara should be doing that, i think that's really for core ecosystem libraries (numpy, pandas) to do, less so for more downstream libraries.

maartenbreddels:
true, but there is a cost for everything. like maintaining workflow files, and we cannot take the responsibility for testing all upstream libraries. What we could do, is to run with --pre on every monday, then at least it's not an extra thing to maintain (our workflow files are complex already)

EwoutH:
Fair. Should I put in in a different file and on a schedule then?

PR #1171, opened 2026-07-04
fix: restructure client reconnect into a supersedable state machine (spurious reloads + empty-mount wedge)

maartenbreddels:
## Summary

The client reconnect path grew several layered fixes this week (1.58.2 + #1167). Before adding another patch, this PR steps back and restructures the ~200 lines of reconnect handling in `main-vuetify.js` into one explicit, documented state machine, and fixes the two remaining production bug classes that the old shape made possible. The verified server-side work from #1167 (comm purge, claim-on-miss fencing) is untouched.

## The two bugs

1. **Spurious teardown of healthy sessions.** The reconnect handler probed `app-status` with a hard one-shot 500&nbsp;ms timeout and treated ANY probe failure as "not started, not recoverable" → `needsRefresh` **and `shutdownKernel` of a healthy kernel**. With `SOLARA_THEME_FORCE_REFRESH=True` (staging config) that is an immediate `location.reload()`; stock production shows the refresh dialog. One late reply (busy backend restoring another kernel, network jitter, or a second reconnect racing the first probe) was enough.
2. **The empty-mount wedge (pre-existing, newly fixed here).** A soft-remount interrupted between `run()` and the mount (socket drop; the `finished` reply is lost) hung forever while holding the `remountInProgress` boolean. Every later remount was silently swallowed, and the next reconnect (probe says `started: true` — the run DID execute server-side) resynced state into a viewless page: empty DOM, no errors, unrecoverable. Ironically, bug 1 sometimes "recovered" these wedges by killing the session.

Shared root cause: reconnect handling had no single owner. Every `connected` event started an independent async flow guarded by ad-hoc flags, and the flows raced each other.

## The state machine (now documented above `connectionStatusChanged` in `main-vuetify.js`)

**State:** `reconnectGeneration` (monotonic; each `connected` event starts a cycle owning its generation — `superseded()` = a newer cycle exists or the socket dropped again), `viewMounted` (true iff a root view is mounted and not torn down since), `needsRefresh` (terminal).

**One cycle = one PROBE, then exactly one verdict, taken only while still the latest live cycle:**

| verdict | condition | action |
|---|---|---|
| PROBE | — | `app-status` on a fresh `solara.control` comm, up to 3 attempts × 2 s with backoff. Only CONSISTENT failure is a verdict. |
| RESUME | `started && viewMounted` | `fetchAll()` (same-instance hot reconnect, unchanged) |
| REPAIR | `started && !viewMounted` | a previous rebuild was interrupted: **re-attach** to the running app — `fetchAll()` + mount the reply's new `containerId` field (the popout attach path). No server state lost. |
| REMOUNT | `!started && canRecover && clientVersion` match | rebuild via `run()` (server restores persisted state) |
| GIVE UP | otherwise | `fallbackToDialog()` — the ONLY place that sets `needsRefresh`/calls `shutdownKernel` |

**Invariants:**
1. At most one cycle — the latest — takes a verdict and touches UI state; there is **no in-flight boolean left that can wedge**. Ownership is the generation number; superseded work aborts silently at its next `await` boundary.
2. A rebuild either completes to a mounted view (`viewMounted = true`) or leaves `viewMounted == false`, which the NEXT cycle detects and REPAIRs. A rebuild hung on a lost reply can never block recovery again.
3. Server half (unchanged, `kernel_context.initialize_virtual_kernel`): a reused context is generation-checked against the backend and superseded BEFORE the websocket is wired, so the probe always observes the post-supersede context.

## Inherited vs redesigned

Inherited from the uncommitted `fix/reconnect-probe-reload` work (rig-verified): the probe retry (3 × 2 s + backoff), the generation counter idea, and the `slow_app_status` integration test. Redesigned on top: `remountInProgress` **removed** (it contradicted generation ownership and was the wedge), `viewMounted` + the REPAIR verdict added (fixes the wedge instead of documenting it), the probe/teardown/rebuild factored into named functions matching the machine's steps, all UI-state writes routed through verdict owners, and the `containerId` field added to the `app-status` reply (server, one line + mypy guard).

## Rig numbers (2 backends + caddy round_robin + redis, `SOLARA_STATE_BACKEND=redis`)

| experiment | master (1.58.2) | this PR |
|---|---|---|
| calm control (6 spaced closes) | 0/1 incidents | 0/1 incidents |
| slow probe (1.0 s app-status latency, one calm close) | **5/5 needsRefresh + kernel shutdown** (session dead) | **0/5**, typing round-trips after |
| forceRefresh parity (staging config, slow probe) | **3/3 full page reloads** | **0/3** |
| rapid bursts (0.2 s latency, 8 bursts of 2–3 drops) | **4/5 runs hit needsRefresh** (bursts 3,4,3,8) | **0/5**, typing OK after every burst |
| designed clientVersion-mismatch reload | 2/2 (works) | 2/2 (preserved) |
| empty-mount wedge (drop mid-remount, run reply lost) | **5/5 dead** (viewless page) | **0/5**, REPAIR re-attaches, typing OK |

## Tests

- NEW `test_interrupted_remount_repairs_on_next_reconnect` (integration, flask+starlette): deterministically loses the remount's `run` reply server-side, red on master (times out — remountCount stays 0 forever), green here; verified the test fails again if the REPAIR branch is neutered.
- NEW `test_app_status_reply_shape` (unit): pins the probe reply fields incl. `containerId`.
- Inherited `test_slow_app_status_probe_does_not_reload` (integration): red on master, green here.
- Gates: `tests/integration/reconnect_state_test.py` + `reconnect_test.py` (flask+starlette) 15 passed; full `tests/unit/` 404 passed; `node --check`; pre-commit (ruff/mypy/codespell) clean.
- `test_reconnect_fail` flake question: 22 runs on unmodified master (isolation + suite context) all green — pre-existing timing sensitivity (the 3 s non-production auto-refresh countdown racing the 5 s contexts-empty wait), most likely load-dependent; not chased here.

## Needs live-staging validation

- The REPAIR path behind the real round-robin LB with redis (rig-verified, but staging has real latency + `SOLARA_THEME_FORCE_REFRESH=True`).
- The probe-retry budget (worst case ~7 s before the dialog on a truly dead-but-socket-accepting server) — acceptable UX?
- Version-skew during a rolling deploy: `started:true` from an old server (no `containerId`) degrades REPAIR to a plain REMOUNT (`run()` — use_state resets, persisted state restored), never a viewless page.

### DIFF
--- a/docs/design-redis-state-persistence.md
+++ b/docs/design-redis-state-persistence.md
@@ -728,19 +728,28 @@ separate triggers:
   genuinely correct** (the browser needs new assets; a soft-remount on stale JS is
   wrong).
 
-`app-status` reply (`app.py:481-489`) gains two fields:
+`app-status` reply (`app.py:481-489`) gains these fields:
 
 ```python
 {"method": "app-status", "started": bool,
+ "containerId": <root container model id when started, else None>,
  "canRecover": <persistence enabled or auto_remount forced, and not recovery-failed>,
  "clientVersion": <opaque asset hash>}
 ```
 
-Client decision on reconnect:
+Client decision on reconnect — one *reconnect cycle* per `connected` event, owned by a
+monotonic generation; a cycle superseded by a newer one (or whose socket dropped again)
+aborts silently, and the probe retries before a failure counts as a verdict (the full
+state machine — PROBE / RESUME / REPAIR / REMOUNT / GIVE UP, with its invariants — is
+documented in the block comment above `connectionStatusChanged` in `main-vuetify.js`):
 
-- `started: true` → today's `fetchAll()` (same-instance hot reconnect — unchanged).
+- `started: true` and the client still has its mounted view → `fetchAll()`
+  (same-instance hot reconnect — unchanged; **RESUME**).
+- `started: true` but the client's view is gone (a previous soft-remount was interrupted
+  mid-way by another socket drop) → **re-attach** to the running app: `fetchAll()` +
+  mount the reply's `containerId` — the popout attach path (**REPAIR**).
 - `started: false && canRecover && clientVersion == own` → **soft-remount** (including
-  after a schema-tag reset — the client can't tell and doesn't need to).
+  after a schema-tag reset — the client can't tell and doesn't need to; **REMOUNT**).
 - otherwise → today's `needsRefresh` dialog: recovery disabled, the client bundle
   changed, or **restore bailed out** (§4.3, `canRecover: false` on a recovery-failed
   context — "we lost", visibly; subject to the bail-out storm valve).
@@ -761,8 +770,12 @@ release; the version-pinned `@widgetti/solara-vuetify-app` bundles are untouched
 
 ### 6.2 Remount sequence (replaces `main-vuetify.js:185-187`)
 
-1. Guard `remountInProgress`; popouts (`?modelid`) keep current behavior (they attach to
-   a specific model that no longer exists; dialog or auto-close).
+1. Ownership by reconnect generation (no in-flight boolean: a boolean guard wedges
+   forever when a rebuild hangs on a reply lost to a socket drop). A rebuild flips
+   `viewMounted` to false first and back to true only after a completed mount, so an
+   interrupted rebuild is *detected and repaired by the next cycle* (REPAIR, §6.1)
+   instead of blocking it. Popouts (`?modelid`) keep current behavior (they attach to a
+   specific model that no longer exists; dialog or auto-close).
 2. Show a light "Restoring session…" indicator (reuse the top progress bar / small
    overlay — not the full loader; keep the stale DOM visible until swap).
 3. Tear down: `widgetManager.clear_state()`/`dispose()` (closes dead comms, disposes
@@ -775,10 +788,11 @@ release; the version-pinned `@widgetti/solara-vuetify-app` bundles are untouched
    `window.location` (pushState routing means the boot-time `path` variable is stale);
    `run(appName, {path, dark, themes})`.
 6. Server restores state before/at first render (Pillars A+B), replies new root
-   `widget_id`; `solaraMount()` swaps it in. Clear the guard and indicator.
+   `widget_id`; `solaraMount()` swaps it in. Set `viewMounted`, clear the indicator.
 
-Edge cases handled: flapping reconnects (guard + re-check connection before final swap;
-last completed attempt wins), fast double reconnect landing back on the original
+Edge cases handled: flapping reconnects (generation ownership: superseded cycles abort
+silently at their next await boundary; the latest cycle repairs whatever an interrupted
+one left behind), fast double reconnect landing back on the original
 instance (server discards its stale context on generation mismatch, §5.1 — the client
 just sees another `started: false` + `canRecover` and soft-remounts; no client-side
 special case needed), multiple tabs sharing one kernel id (each re-runs `run`; restore
--- a/solara/server/app.py
+++ b/solara/server/app.py
@@ -544,6 +544,11 @@ def solara_comm_target(comm, msg_first):
             reply = {
                 "method": "app-status",
                 "started": started,
+                # the root container's model id. A client whose rebuild was interrupted (socket
+                # drop between run() and the mount) has no view left but finds started=true here
+                # on its next reconnect; this id lets it RE-ATTACH to the running app (fetchAll +
+                # mount, the popout path) instead of resyncing state into a viewless page.
+                "containerId": context.container._model_id if context.container is not None else None,
                 # whether the client may soft-remount instead of showing the refresh dialog (§6.1)
                 "canRecover": _can_recover(context),
                 # opaque asset hash; a mismatch with the client's baked-in solara.clientVersion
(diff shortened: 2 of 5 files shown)


maartenbreddels:
## Independent local browser validation (A/B master vs branch)

Validator: an independent Claude Code agent (model id: `claude-fable-5`), driving a real (headed) Chrome via CDP — not the PR author's Playwright rig. No code was changed; both sides ran unmodified checkouts.

**Rig**: 2 single-worker solara servers (ports 8801/8802, `--production`, `SOLARA_STATE_BACKEND=redis` on a real redis db 7, shared) behind caddy `lb_policy round_robin` (:8800). A/B via two detached `git worktree`s — master @ `59b83a31` (1.58.2) and this PR @ `b7b9f902` — selected with `PYTHONPATH`; the served `main-vuetify.js` hash was verified against the worktree file at runtime on both nodes, and the running page was fingerprinted (`solara.debug.viewMounted` getter present only on the PR build). Test app: InputText bound to a reactive (`TEXTLEN:n` echo proves the server round trip), enable-gated button, one persisted reactive. Failure predicates: `window.__rigBootT` reload canary + `performance.navigation` type, `app.$data.needsRefresh`, dialog text, and typing round-trips (fresh marker length each probe, so a stale DOM cannot false-pass).

### Verdict table

| scenario | master (1.58.2) | this PR |
|---|---|---|
| 1. baseline: single calm close | OK — RESUME, persisted value restored, typing syncs | OK — cross-node REMOUNT, persisted `CLICKS` restored from redis, typing syncs |
| 2. slow probe (`app-status` +1.0 s, `SOLARA_THEME_FORCE_REFRESH=True`), 1 calm close on warm nodes | **3/3 actual `location.reload()`** (canary gone, `navigation.type == "reload"`, session state wiped) | **0/3 (0/4 incl. warm-up close)** — no reload, no `needsRefresh`, typing syncs after every close |
| 3. rapid bursts (+0.2 s latency, 2–3 closes @ 0.5–1.5 s, up to 8 bursts/run) | **2/2 runs died with `needsRefresh`** (run 1: burst 4, run 2: burst 8) — "Could not restore session" dialog on a healthy session | **0/16 bursts across 2 runs** — `remountCount` == `reconnectCount` advanced sanely (up to 22), typing OK after every burst, 0 page errors, 0 `Comm is already created` |
| 4. empty-mount wedge (`run()` slowed 1.5 s, socket cut mid-remount) | **3/3 dead viewless sessions**: input gone (body ≈ 46 chars), `remounting` stuck `true`, `remountCount` 0, `needsRefresh` false (no dialog, no errors), and an extra reconnect does NOT recover (stuck on "Restoring session…") | **4/4 recovered**: 3 recovered by the next cycle's REMOUNT (no wedged boolean), 1 explicitly exercised **REPAIR** — probe reply `started:true` + `containerId` while `viewMounted:false`, view re-attached, typing syncs |
| 5. designed clientVersion-mismatch refresh (PR side) | n/a | **2/2 preserved** — faked `solara.clientVersion` → refresh dialog (`needsRefresh`) on the first cross-node close |

### Anomalies / notes for reviewers

- **Background-tab timer throttling confound (validation methodology, but also a real-world observation):** with the tab hidden, Chrome throttles master's one-shot 500 ms probe timeout to ~1 s alignment, so master's slow-probe bug becomes a coin flip against a 1.0 s reply. All counted runs above were taken with the tab visible and focused (matching the PR rig's active Playwright page). Corollary: in production, master's spurious-reload rate depends on tab visibility.
- **Rig knob cold-start hole:** `RIG_SLOW_APP_STATUS` installs its wrap when the app module first executes in a server process; a node that has never run the app replies fast. The first cross-node close after a rig start can therefore remount cleanly even on master. Both nodes were warmed (and per-node latency measured at ~1.0 s via a manual 5 s-timeout probe) before any counted attempt — this is also why "close after ≥1 remount" matters when reproducing.
- **REPAIR needs round-robin parity to be exercised:** in 3/4 wedge attempts the post-interrupt reconnect landed on the *other* node (`started:false` on both probes in the server logs) and recovery was a clean REMOUNT — still a pass, but it exercises the generation ownership fix, not REPAIR. Inserting one extra HTTP request after the interrupt shifted the parity and produced a textbook REPAIR trace: `reply{started:true, containerId:13a0755a…}` at `viewMounted:false` → view re-attached, `remountCount` +1, typing round-trips.
- Master's wedge is *worse* than "one dead page": the stuck `remountInProgress`/`remounting` flag means every later reconnect resyncs silently into the viewless page (`connected`, no dialog, no console errors) — unrecoverable without a manual refresh, exactly as the PR body claims.
- Screenshots captured: master's spurious "Please refresh the page / Could not restore session" dialog over a healthy session (bursts), master's blank wedged page, PR's healthy post-REPAIR page.

**Verdict: the PR's claims reproduce cleanly in a real browser.** Master exhibits all three failure classes (spurious reload on one slow probe: 3/3; burst-death: 2/2 runs; unrecoverable empty-mount wedge: 3/3). The PR branch shows zero incidents in the same scenarios (0/3+0/16+0/4), the REPAIR verdict works as designed behind real round-robin + redis, and the designed bundle-mismatch refresh is preserved (2/2).

model id: `claude-fable-5`

PR #1173, opened 2026-07-04
ci: parallelize integration tests and speed up installs

maartenbreddels:
## Summary

CI wall-clock is dominated by `integration-test (windows, 3.9, 7)`: ~17 min, of which ~12 min is the pytest step running playwright tests serially on a 4-core runner (measured on run 28710308256). This PR targets that critical path:

- **`-n 2` (pytest-xdist) for `tests/integration`.** The port schemes in `tests/integration/conftest.py` and `solara/test/pytest_plugin.py` already referenced `PYTEST_XDIST_WORKER`, but spaced workers **1 port apart** while each worker starts 2–3 servers — concurrent workers would collide. Now spaced 3 (integration conftest) and 10 (plugin) apart. Two workers is the max that keeps all ports inside the valid auth0 callback range (18765–18770) for the oauth tests — noted in a comment next to `-n 2`.
- **One pip invocation instead of six.** Single resolve, no install-then-downgrade churn; the windows install step was 2.8 min. `pytest-xdist` is installed explicitly in the locked path until the next lock refresh pins it.
- **`playwright install chromium`** — the tests only use chromium; skip downloading firefox/webkit (also for the pyinstaller job).
- **`PYTHONUNBUFFERED=1`** workflow-wide: force-exiting processes (`os._exit` on windows, pytest-timeout kills) can no longer lose buffered output, and logs stream in real time.

Expected result: integration windows ~17 → ~10-11 min, ubuntu ~12 → ~7 min; total CI wall-clock from ~18-19 min to ~12 min.

Validated locally (macOS): `-n 2` on a shifted port range with both flask+starlette session servers per worker — no port conflicts.

Sibling PR with the flake fixes: https://github.com/widgetti/solara/pull/1172 — independent, but both reduce the "rerun CI to release" pain.

## Test plan
- [x] local xdist smoke test (2 workers, flask+starlette servers each)
- [x] pre-commit green
- [ ] full CI matrix on this PR (the real validation)
- [ ] compare job timings against master baseline run 28710308256


maartenbreddels:
Round-1 CI exposed two deeper issues, both fixed in follow-up commits:

1. **Port drift** (`4045b39`): with xdist load scheduling, the parameterized session `solara_server` fixture is torn down and re-created on every flask↔starlette param switch, so the `TEST_PORT += 1` counter drifted into the other worker's range and past the valid auth0 callback ports (a server was observed on 18771 → the oauth login form never appeared). Ports are now a pure function of (worker, param), and the one-off servers in cmdline/flask/starlette tests use OS-assigned free ports.

2. **pytest-retry × tmp_path** (`81ef288`): pytest-retry (≤1.7.0, the latest) with pytest 8 makes any *retried* test that uses `tmp_path` error in teardown with `KeyError: <StashKey>` — a successfully retried flaky test turns into a hard failure. Reproduces with 3 lines locally, **without xdist**, so this can already bite master today (integration tests use `--retries 3`); parallel windows runs just retry more often. Worked around by seeding the stash entry in a teardown hookwrapper — can be removed once fixed upstream.

Remaining known flake on this branch: `test_cdn_via_altair` (the CDN-proxy 500, fixed in #1172) — will disappear after #1172 merges and this branch is rebased.

Timing so far (vs master baseline run 28710308256): integration ubuntu test step ~12 min → ~7 min, windows job 17.1 min → ~14 min.

maartenbreddels:
Final round: **34/34 green**, no reruns needed.

Timings vs the master baseline (run 28710308256):

| Job | before | after |
|---|---|---|
| integration windows | 17.1 / 16.5 min | 12.1 / 11.8 min |
| integration ubuntu | 12.0 / 9.2 min | 9.2 / 7.3 min |
| pyinstaller windows | 7.0 min | 4.6 min |
| unit windows | 4.5–4.9 min | 3.4–3.5 min |
| **total wall clock** | **~19 min** | **~14 min** |

The lock-refresh bot already committed the regenerated unit locks on this branch (bqplot 0.12.47, pywinpty 2.0.13 wheel, pytest-asyncio in, tornasync out), so the locked install path is proven too.

PR #1176, opened 2026-07-05
fix: free closed virtual kernels promptly instead of leaving them to the gc

maartenbreddels:
## Summary

Closed virtual kernel contexts were unreachable but stuck in **reference cycles**, so refcounting could not free them — they waited for a generation-2 `gc`, which under load shows as a memory sawtooth whose peaks scale with per-kernel state. Measured with a small task-heavy test app (10 pages x 10 cycles, real Chromium sessions): idle footprint bounced up to **231 MB**; after this PR it converges to a smooth **~170 MB plateau, peak 174 MB**. A companion docs PR adds the measurement handbook + harness used to find all of this.

## The cycles (found with a gc-disabled weakref diagnostic) and their fixes

1. **Kernel-cull task** (affects every app): `close()` cancelled `_last_kernel_cull_task` but kept the reference. A cancelled asyncio task retains its `CancelledError`, whose traceback holds the `kernel_cull` frame, whose closure holds the context: `context → task → traceback → frame → context`. `close()` now drops the task/future references after cancelling.
2. **Task bookkeeping**: `TaskAsyncio.current_task` keeps the finished/cancelled `asyncio.Task`, which pins the kernel context via its contextvars copy and exception traceback. A `context.on_close` callback now clears the call state (`_drop_call_state`), and `is_current()` returns `False` for stale calls after close.
3. **`app_object`**: `close()` closed the render context but never dropped the reference.

## Related fixes the investigation surfaced

- **Future delivery race**: a task finishing while the page closes could call `future.set_exception()` on an already-done future → `InvalidStateError` raised inside the loop callback → asyncio logs it *with the exception traceback attached*, recreating the frame-pinning cycle above (and log noise). Deliveries now go through a guarded threadsafe helper that checks `future.done()` inside the loop callback. Cancellation is delivered as `future.cancel()` so no exception object crosses the thread boundary. Red→green unit test included (`test_task_finishing_after_future_cancelled_is_not_an_error`).
- **Delivery closure bug**: the new delivery lambdas must not capture the `except ... as e` name (unbound when the block exits, the lambda runs later on the loop) — covered by `test_task_async_future_receives_exception`.
- **fd leak**: the per-call thread event loop in `TaskAsyncio` was never closed, leaking its selector fd until collected.
- **`Thread.start()` hang**: a thread started while its inherited kernel context closes (`kernel` already `None`) crashed solara's patched bootstrap *before* `_started` was set, hanging the caller forever. The bootstrap now degrades to no-context.

## What remains (upstream)

One cycle lives in reacton: `make_setter.<locals>.set_` closures capture the render context strongly and are stored in hook state and widget callbacks, so the render tree (which owns `use_memo` payloads) stays cyclic after `rc.close()`. Until that is fixed upstream, the new `kernel.gc_after_close` setting (default on, `SOLARA_KERNEL_GC_AFTER_CLOSE=false` to disable) schedules a **deferred, coalesced** `gc.collect()` ~1s after a kernel closes — off the close path, one collect per burst of closes. Measured: disabling it brings the sawtooth back; with it the plateau holds. It also covers cycles in user code.

## Testing

- `tests/unit/task_test.py`: two new tests (one red on master by design), full unit suite green (410 passed).
- Object-level leak tests (weakref on context/kernel/shell after page close) pass repeatedly on flask + starlette, with `REACTON_FAST=1` too (test lands in the docs PR alongside its test app).
- Memory benchmarks before/after in the companion docs PR (`docs/memory-usage-inspection.md`).

### DIFF
--- a/solara/server/kernel_context.py
+++ b/solara/server/kernel_context.py
@@ -234,8 +234,14 @@ class VirtualKernelContext:
                 self.page_status[key] = PageStatus.CLOSED
             if self._last_kernel_cull_task:
                 self._last_kernel_cull_task.cancel()
+                # drop our reference: a cancelled task keeps its CancelledError whose
+                # traceback holds the kernel_cull frame, whose closure holds `self` -
+                # keeping the reference would turn that into a reference cycle through
+                # the whole context, which refcounting cannot free
+                self._last_kernel_cull_task = None
             if self._last_kernel_cull_future:
                 self._last_kernel_cull_future.cancel()
+                self._last_kernel_cull_future = None
             if self.closed_event.is_set():
                 logger.error("Tried to close a kernel context that is already closed: %s", self.id)
                 return
@@ -249,6 +255,9 @@ class VirtualKernelContext:
                     except Exception as e:
                         logger.exception("Could not close render context: %s", e)
                         # we want to continue, so we at least close all widgets
+                # drop the reference: anything that still (indirectly) references this
+                # closed context should not keep the whole render tree alive with it
+                self.app_object = None
             widgets.Widget.close_all()
             # what if we reference each other
             # import gc
@@ -269,6 +278,8 @@ class VirtualKernelContext:
                 if _ctx is self:
                     del current_context[key]
             self.closed_event.set()
+        if solara.server.settings.kernel.gc_after_close:
+            _schedule_gc_after_kernel_close()
 
     def _state_reset(self):
         state_directory = Path(".") / "states"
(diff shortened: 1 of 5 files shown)


maartenbreddels:
The `test_kernel_lifecycle_double_disconnect` failure on ubuntu/3.8: `gc_after_close` scheduled a full collection after *every* unit test (each closes its fixture kernel context), and on the loaded 2-core runner the GIL pauses stretched the test's 0.1s sleep past the 0.2s cull window, so the reconnect hit an already-culled kernel. Fixed by only scheduling the gc for kernels that actually rendered an app — a bare context holds nothing worth collecting. Full unit suite green locally.

PR #1185, opened 2026-07-06
fix: a task result write racing the use_task unmount cleanup re-pinned the result

maartenbreddels:
## Summary

`test_use_task_unmounted_instance_is_collected` failed once on master (`macos, 3.8, 8.0`, solara-2.0-mode run: `2/6 payloads pinned` after the test's 10 s gc loop) while the identical SHA passed in another run. Hunting it with an instrumented retainer dump under CPU contention (~4% local reproduction) caught the chain red-handed — and it's a real production leak path in the #1180 unmount cleanup, not test noise.

## The race

`context.user_dicts → StoreValue → TaskResult → payload`, rooted: a worker thread passes its `is_current()` check, the GIL switches, the unmount cleanup drops the call state and clears the kernel-store entries, then the worker completes its already-decided `self._result.value = ...` write — **re-creating the cleared entry with the payload inside**. Nothing clears it again until kernel close; no gc can help (it's rooted).

**Fix — lock-free, last-writer-loses**: `_drop_call_state()` sets `_call_state_dropped` *before* the cleanup clears; all result writes go through `Task._set_result()`, which re-checks the flag *after* writing and clears its own write. Every interleaving converges to cleared, no lock on the hot path, idempotent with the cleanup's clear.

## Second bug the same dump exposed

The cleanup read `task_instance._result` *through the Proxy* after clearing the singleton store entry — `Proxy.__getattr__` → `Singleton.get()` → the factory **re-created the entry just cleared** (a fresh `TaskAsyncio` pinned until kernel close). The cleanup now resolves the instance once up front, operates on it directly, and clears the singleton entry **last**.

## Test hardening (the second, benign failure mode)

Under 4-way contention one clean run still took 11.45 s: the test's own gc churn starves the tiny worker threads (delayed collection, not a leak — the payload is freed as soon as the thread runs). Deadline 10 → 30 s, gentler loop sleep. The test still exits in <1 s when clean.

## Verification

- Reproduction: 1/25 runs exceeded 10 s before; retainer dump caught the rooted chain (in the PR description above).
- With the fix: **120/120 contended runs** (4 parallel × 30) in solara-2.0 mode, max 11.45 s, plus full suite green in both modes (419 + 421 passed).

### DIFF
--- a/solara/tasks.py
+++ b/solara/tasks.py
@@ -203,7 +203,25 @@ class Task(Generic[P, R], abc.ABC):
     def is_current(self) -> bool: ...
 
     def _prestart(self):
-        self._result.value = TaskResult[R](latest=self._last_value, _state=TaskState.STARTING)
+        self._set_result(TaskResult[R](latest=self._last_value, _state=TaskState.STARTING))
+
+    # set by _drop_call_state (use_task component unmount); a worker thread that already
+    # passed its is_current() check must not resurrect the cleared store entry
+    _call_state_dropped = False
+
+    def _set_result(self, result: "TaskResult[R]") -> None:
+        """Write the task result, without racing a use_task unmount cleanup.
+
+        The unmount cleanup clears this task's kernel-store entries, but a worker thread
+        that passed its is_current() check just before the cleanup ran would re-create the
+        entry with its write - pinning the result (and everything it references) until the
+        kernel closes. Re-checking after the write makes the last writer lose in every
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Superseded by #1186, which now targets master and carries both commits: the race analysis + hardened test from this PR, with the mechanism replaced by the component-scoped result store (the band-aid machinery is deleted rather than maintained).

PR #1190, opened 2026-07-08
Closed AutoSubscribeContextManager must not resubscribe

maartenbreddels:
## TLDR

Follow-up to #1189: four fixes closing every path by which per-kernel subscriptions outlive their kernel. Found and validated against a large production application with breadcrumb-instrumented listener entries (creator thread, time, kernel state, call site) — each fix has a deterministic test that fails on the commit before it.

1. **Closed managers must not resubscribe** — a recompute racing `unsubscribe_all` (task threads, teardown fires) re-created subscriptions on the dead scope. Flip-flag-then-clear + recheck-after-write.
2. **`on_close` on an already-closed context runs the callback immediately** — lazy factories re-create per-kernel resources after close; their cleanup registration used to vanish.
3. **`on_close` registrations arriving mid-drain must not be lost** — `reversed()` over the live list skipped late arrivals; while+pop double-drain, deliberately lock-free (a locked variant AB-BA-deadlocked closes against store init locks: 22 stuck kernels in the validation rig).
4. **Lock-free commit-chain compute state for `AutoSubscribeContextManager`** — the manager is shared by every recomputing thread; in-place `self.subscribed = {}` orphaned a concurrent thread's unsubscribe closures (production: login task vs render, ~1% of kernels, a whole computed-chain each). In-flight state is now per-thread; commits swap the committed dict under a microsecond leaf lock and diff, so concurrent commits clean up after each other. `test_computed_overlapping_recompute_no_orphan` reproduces the production interleaving deterministically.

Also: case study 4 + epilogue in `docs/memory-usage-inspection.md` (the counter-census methodology that found these), the amplification canary app, and the toestand suite now runs with server mode active so kernel-close paths are actually exercised.

## Validation

- Rig (production app, real data): stepper events 3-in-11-cycles → **0-in-12**, residue frozen; 42-cycle stability run decays to ~1–2 MB/cycle floor (−86% vs pre-#1189 baseline).
- Unit: every fix fails on its predecessor commit; full suite 425 passed; mutation-detection mode green; mypy clean.

### DIFF
--- a/docs/memory-usage-inspection.md
+++ b/docs/memory-usage-inspection.md
@@ -637,6 +637,34 @@ render context's effect cleanups, so cleanups must tolerate
 already-removed entries (discard semantics), or the purge must run after
 teardown — a raising cleanup aborts the whole render-context close.
 
+**Epilogue: the fix itself raced — twice.** With the owner cleanup shipped,
+a 25-cycle run of the same production app still leaked a constant
+~10 MB/cycle. The counters (not the memory numbers) named it: after gc with
+zero kernels, one surviving `AutoSubscribeContextManager` per (Computed,
+kernel) and their listener entries were back. Two lessons in reading that
+census:
+
+- **Subclass discrimination**: the Reacton (component) managers counted 0 —
+  their effect-cleanup path works — all survivors were the plain (Computed)
+  kind. One count split the search space in half.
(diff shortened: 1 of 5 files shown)


maartenbreddels:
Superseded by #1192, which rebuilds the fix from a clean baseline with consolidated memory-leak, restart/close race, and deadlock regression coverage.

Issue #1093, opened 2025-08-20
how can I clear the memory after a session is closed?


dfguerrerom:
When testing the memory usage of a Solara application (calculator example) deployed inside a Docker container, the container’s memory consumption increases each time a new client connection is opened. However, the memory is not released after the sessions are closed.

## Steps to Reproduce
- Deploy the calculator example inside a Docker container.
- Monitor container memory usage.
- Open multiple client connections and observe memory increase.
- Close the connections.

## Expected Behavior
Memory usage should decrease once sessions are closed (especially when cull_timeout is set to 0).

## Actual Behavior
Memory usage increases with each new session but never decreases, even after all sessions are closed.

## Additional Information

cull_timeout has been explicitly set to 0 seconds.

<img width="947" height="780" alt="Image" src="https://github.com/user-attachments/assets/5a0c181d-a5c4-49ef-a672-9a25fd1f666e" />

PR #1192, opened 2026-07-10
fix: make reactive subscription lifecycle leak-free

maartenbreddels:
## Summary

Fix reactive subscription and virtual-kernel lifecycle leaks without introducing callback/lock deadlocks.

This replaces #1190 with a clean implementation built from the strongest regression tests collected across the earlier branches.

## Root cause

Subscriptions, automatic dependency tracking, and kernel restart/close cleanup had overlapping ownership:

- transient Reactive, Singleton, and Computed state could remain in kernel storage;
- automatic subscriptions could survive their owner or resubscribe after terminal close;
- listener and leak-check bookkeeping could accumulate empty entries or callbacks;
- restart and close callbacks ran with lifecycle state and locks that allowed races or re-entrant deadlocks.

## Changes

- make kernel restart/close state transitions explicit and run user cleanup outside lifecycle locks;
- make on-close registration safe during and after close;
- weakly track kernel store ownership while preserving module state and shared explicit keys across hot reload;
- make automatic subscriptions per-run, terminally closeable, and safe for overlapping or failed runs;
- prune listener and leak-check bookkeeping when cleanup completes;
- add focused regression coverage for leaks, repeated restart, close races, callback failures, explicit-key collisions, and destructor re-entry.

## Validation

- 454 unit tests passed, 23 skipped;
- 4 server/browser memory-leak integration tests passed;
- 21 lifecycle, reconnect, and ESM integration tests passed, 5 skipped;
- pre-commit, Ruff, formatting, mypy, and codespell passed;
- independent memory, deadlock/race, and simplicity reviews all returned GO;
- concurrency stress passed 30 iterations.

### DIFF
--- a/solara/server/kernel_context.py
+++ b/solara/server/kernel_context.py
@@ -18,6 +18,7 @@ import re
 import threading
 import time
 import typing
+import weakref
 from pathlib import Path
 from typing import Any, Callable, DefaultDict, Dict, List, Optional, Tuple, Union, cast
 
@@ -52,6 +53,13 @@ else:
 class PageStatus(enum.Enum):
     CONNECTED = "connected"
     DISCONNECTED = "disconnected"
+    CLOSED = "closed"
+
+
+class _LifecycleState(enum.Enum):
+    OPEN = "open"
+    RESTARTING = "restarting"
+    CLOSING = "closing"
     CLOSED = "closed"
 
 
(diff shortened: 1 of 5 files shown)


dfguerrerom:
might this be related to https://github.com/widgetti/solara/issues/1093 ?

PR #1198, opened 2026-09-03
Make the router effect run once and log uncaught exceptions with exc_info

maartenbreddels:
Two changes from a production incident (reacton 1.10.3, https://github.com/widgetti/reacton/pull/54).

## Router effect

`RoutingProvider` re-queued `solara.get_widget(main)` in an effect without dependencies, on every render of the router. When a render pass aborted halfway, reacton before 1.10.3 could do that, the chained closure held an unreconciled `main`, `get_widget` raised from inside the effect, and the whole app was replaced by the traceback. reacton 1.10.3 fixes the abort; this makes the router not depend on it.

- `solara.use_effect(get_nav_widget, [])`: the root element of the component is always the same `VBox` at the same slot, reacton updates it in place, so the Navigator widget is the same object for the lifetime of the app. The `nav_widget.current.location = path` line still runs on every render.
- `get_widget(nav)` instead of `get_widget(main).children[0]`. The old comment said it did not work; it does on reacton 1.10.3: `nav` is a widget element reconciled in the VBox's child context, which the downward search from the router context reaches. The stale comment is gone.
- Test: `test_routing_provider_navigator_widget_stable` pushes three paths and one browser-side location change, asserts the widget identity is stable and `location` follows every push. Rendered with `handle_error=False`, so a failing effect would raise.

## Uncaught exception logging

`FakeIPython.showtraceback` logged `"Uncaught exception: %s"` with the formatted traceback in the message and no exception info. Error trackers group on the message, so every escaped event-handler exception of every app lands in one issue. Now `logger.error("Uncaught exception", exc_info=value)`; the pdb branch and the control-socket message with the formatted traceback are unchanged. Test asserts the record carries the exception.

### DIFF
--- a/solara/autorouting.py
+++ b/solara/autorouting.py
@@ -115,15 +115,24 @@ def RoutingProvider(children: List[reacton.core.Element] = [], routes: List[sola
     solara.routing.router_context.provide(solara.routing.Router(path, routes=routes, set_path=set_path_with_redirect))
 
     def get_nav_widget():
-        # not sure why get_widget(nav) does not work
-        nav_widget.current = solara.get_widget(main).children[0]  # type: ignore
+        # get_widget(nav) did not work when this was written (2023, "not sure why"): reacton's
+        # get_widget only searched the current component context, and nav is reconciled in the
+        # VBox component's child context. reacton 1.7.2 (dfdca66) made it search child contexts,
+        # and solara requires reacton>=1.9, so the detour via get_widget(main).children[0] is gone.
+        nav_widget.current = solara.get_widget(nav)  # type: ignore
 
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Why `get_widget(nav)` works now, and why it did not when the comment was written in 2023 (#81): reacton's `get_widget` only searched the current component context then, and `nav` is reconciled inside the `VBox` component's child context, hence the detour via `get_widget(main).children[0]`. reacton dfdca66, "get_widget did not always work for function components", released in v1.7.2, made it search child contexts. solara requires `reacton>=1.9`, so the detour is no longer needed on any supported reacton. The code comment now records this.

